PR #85, opened 2026-07-26
fix: renumber duplicate migration 000077_ai_variables_resolved to 000078 so the backend boots

Turtlesfr:
## Problem

Current `main` ships two migrations with the same version number:

- `000077_warmup_generation_scale.{up,down}.sql` (merged via #82)
- `000077_ai_variables_resolved.{up,down}.sql` (merged via #83)

golang-migrate refuses duplicate versions, so a fresh `docker compose up --build` crash-loops the backend on boot:

```
Failed to run migrations: failed to create migration source: failed to init driver with path migrations: duplicate migration file: 000077_warmup_generation_scale.down.sql
```

## Fix

Renumber `000077_ai_variables_resolved` → `000078_ai_variables_resolved` (rename only, contents untouched).

Renumbering the ai_variables pair rather than the warmup pair matches merge order: #82 (warmup) landed on main before #83 (ai_variables), so any deployment that migrated between the two merges has already recorded version 77 = warmup_generation_scale. With this rename such deployments apply 000078 cleanly on next boot; fresh installs get both in order.

No Go references to the migration filenames exist (`ai_variables_resolved` mentions in `pg_campaign_progress.go` are the column name). Verified locally: backend boots, migrations apply, seeded stack runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


meszmate:
Oh, thank you. I wasn’t paying enough attention.

PR #87, opened 2026-07-31
Fix unibox NEW_EMAIL ingestion

PattrnData:
## Summary
- wrap worker NEW_EMAIL events with user_id and message payload
- accept legacy bare NEW_EMAIL payloads by recovering user_id from the email account
- guard nil NEW_EMAIL messages and normalize nil recipient arrays before unibox insert

## Test Plan
- go test ./internal/repository ./internal/app/worker/wmail ./internal/app/consumer
- docker compose build consumer worker on mailserver
- live controlled Graph send/reply inserted into unibox_emails


PattrnData:
Closing this PR intentionally. This Warmbly work is a custom branch/live-mailserver lane and should not be proposed for merge into main. Keeping the branch/commits as evidence/reference only.

PR #95, opened 2026-08-09
feat: IMAP EXISTS/UIDNEXT fallback for Hostinger CONDSTORE stall

tjsasakifln:
## Summary
- LIST-STATUS now requests EXISTS + UIDNEXT alongside HighestModSeq
- When MODSEQ is flat but EXISTS/UIDNEXT advances (Hostinger CONDSTORE stall), fetch the new sequence window with ChangedSince 0
- Pure `planMailboxSync` unit tests cover first sync, no-op, CONDSTORE advance, Hostinger stall, and UIDNEXT-only paths

## Why
Continuous 1m IMAP poll missed Hostinger self-replies after the first full sync because `sync_imap` only called `FetchChanges` when `HighestModSeq` advanced. Reply-stop then depended on operator `ADD_EMAIL` reseed. That is unsafe for pilot cadence (OBJECTIVE §26).

## Test plan
- [x] `go test ./internal/app/worker/wmail/`
- [x] `make lint`
- [ ] Deploy worker to CONFENGE VPS; after first attach, send natural Re: and prove Unibox + PLANNED→REPLIED within continuous poll (no backend restart, no JetStream force)


tjsasakifln:
Opened on wrong remote; tracking https://github.com/tjsasakifln/warmbly/pull/31

PR #240, opened 2026-08-29
fix: record why a warmup generation batch failed, instead of only 'batch failed'

joaoppa:
A warmup generation batch that fails as a whole records `batch failed` and nothing else, discarding the reason the provider gave.

Observed on a self-hosted install: three consecutive scheduled runs failed over about 13 hours, and the admin panel said only:

```
28/08 18:55  failed  batch failed
28/08 12:25  failed  batch failed
28/08 05:55  failed  batch failed
```

The actual reason was sitting on the batch object the whole time:

```json
"errors": {"data": [{"code": "invalid_request",
  "message": "Cannot find file file-…, or organization org-… does not have access to it."}]}
```

(That message turned out to be OpenAI's way of reporting a monthly spend cap: the file existed and was readable with the same key throughout. Misleading or not, it is what the operator needed to see, and it was the only thing pointing anywhere.)

## Why the existing path does not cover this

`ingestBatch` already falls back to the error file when a *completed* batch has no output, which covers a batch whose requests were all refused. A batch with status `failed` is different: it never ran, so there is no error file at all, and `pollBatchJob` writes the status alone:

```go
case "failed", "expired", "cancelled":
    if job.Error == "" {
        job.Error = fmt.Sprintf("batch %s", status)
    }
```

## Change

`GetBatch` returns a `BatchState` instead of a widening list of strings (it was already at five returns and this would have made six), carrying the batch-level `errors` alongside the status, files and counts. The failed branch appends the provider's message to the status it already records.

`batchFailureReason` renders one line and names how many more there are rather than pasting every entry into a job row, skips blank messages, and returns empty when there is nothing to say, so the existing `batch failed` text stands unchanged when the provider offers no reason.

## Tests

`TestBatchFailureReason` covers no errors, one error, several (the "+N more" form), an entry with a blank message ahead of a real one, and all-blank. It fails on the code before this change, since the function did not exist.

The poller branch itself is not unit-tested here: `warmupcontent.service` holds a concrete `*generation.GenerationClient` and `Enabled()` is `gen != nil`, so stubbing it means introducing an interface and rewiring the constructor, which is more than this fix should carry. The wiring is two lines and was checked against the recorded failure above.

`gofmt -l` clean, `go build ./...`, `golangci-lint v1.64.8 ./internal/...`, `scripts/check-migrations.sh` and `go test -race -count=2` on the two touched packages all clean.

### DIFF
--- a/internal/app/warmupcontent/batch.go
+++ b/internal/app/warmupcontent/batch.go
@@ -168,29 +168,35 @@ func (s *service) PollBatches(ctx context.Context) error {
 
 // pollBatchJob reconciles a single batch job.
 func (s *service) pollBatchJob(ctx context.Context, job *models.WarmupGenerationJob) error {
-	status, outputFileID, errorFileID, counts, err := s.gen.GetBatch(ctx, job.BatchID)
+	state, err := s.gen.GetBatch(ctx, job.BatchID)
 	if err != nil {
 		return err
 	}
-	job.BatchStatus = status
+	job.BatchStatus = state.Status
 
-	switch status {
+	switch state.Status {
 	case "completed":
-		job.BatchOutputFileID = outputFileID
+		job.BatchOutputFileID = state.OutputFileID
 		// A batch whose requests all failed completes with no output file; the
 		// refusals are in the error file, which has the same JSONL shape. Read
 		// it instead of failing on the empty id, or the reason is never seen.
-		resultsFileID := outputFileID
+		resultsFileID := state.OutputFileID
 		if resultsFileID == "" {
-			resultsFileID = errorFileID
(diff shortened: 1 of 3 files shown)


meszmate:
Pushed a commit. The `(+N more)` suffix was sized off `len(errs)` but blank entries get skipped, so `[{Code: "empty"}, {Message: "the real one"}]` claimed a second reason that isn't there. It counts readable messages now. I retargeted your blank-message case and added one to pin the count.

Also shortened the new comments for Greptile's P2.

Lint and tests clean. Good catch on the underlying bug, the writeup made this easy to verify.

PR #248, opened 2026-08-29
feat: make the campaign content check actually fire

### LINKED ISSUE #144: [Campaigns] P1 - Campaign-body content lint as pre-send and editor check (extend warmlint)
**Priority: P1 (high).**

## Summary
`warmlint.Check` runs only on warmup content. Campaign subject/body are sent with no spam-trigger scoring, no link-count cap, no ALL-CAPS or attachment checks, so obvious spam copy ships with zero feedback.

## Proposal
- generalize `internal/pkg/warmlint` into a scoring function returning findings + severity, adding link/attachment/image heuristics;
- surface findings in the sequence editor at save time (additive response field);
- soft-gate at `campaign_task.go` after `expandSpintax`: log `content_warning`; trigger the breaker path for hard cases. Advisory by default behind an existing-style Preflight setting.

## Importance alignment
- **Mox:** content quality feeds its Bayesian junk learning and delivery acceptance, i.e. content signals are part of the system, not decoration.
- **Notifuse:** its editor is a core differentiator (MJML + preview); template-quality feedback at save time is table stakes in that product family.

## Reference
`resources/warmup-abuse-cold-research.md` section 3.3.

meszmate:
Follow-up to #229, which closed #144. Verifying that work end to end turned up six defects that meant the content check largely did not do what the issue asked for.

## What was broken

**The link cap never fired.** `Score` counted `https?://` in the plain-text body, but an HTML email carries its URLs in `href` attributes, and both `stripTags` and the editor's `htmlToPlain` discard them. A body with six anchors scored a clean 100/100. The link-count cap is one of the three heuristics #144 named explicitly, and it was dead on the common case. It now counts hrefs and bare URLs and takes the larger, so a URL used as its own anchor text is not counted twice.

**Every campaign with a wait or action step failed preflight.** `GetSequencesByCampaignID` never selected the `kind` column, so `contentScoreCheck` scored control nodes as copy. An empty subject plus an empty body scores 55/100, under the default floor of 60, so the launch dialog reported `Step 2 scores 55/100 for spam signals` about a step that was never an email. Fixed at both levels: the query selects `kind`, and scoring skips non-email steps.

**The one warning-level log in the system rendered grey.** `content_gate.go` wrote `"level": "warning"`; `getCampaignLogs` maps `metadata.level` and the feed only tints amber on `"warn"`.

**Preflight ignored attachments while the send path scored them**, so the launch dialog could report 100/100 and the activity feed then contradict it. The advanced service now takes an optional attachment repository, with a compile-time assertion so the type-assertion wiring in `main` cannot silently become a no-op.

**`min_content_score` had no server-side validation.** An API caller could store `5000` and flag every campaign forever. Clamped to 1-100 on write; out-of-range legacy rows fall back to the default on read.

**The docs claimed the editor scores live as you write.** It was a manual button. The editor now re-scores on the same debounce the composer's preview already uses, and the redundant button is gone.

Also in here: the em dashes are out of this feature's user-facing copy (five of them in one small feature), and the send-path gate scores before reading settings, so clean copy no longer costs two database queries per recipient.

## Verification

`gofmt` clean, `make lint` clean, `pnpm typecheck` and `lint` clean in `web/`, `types:check` / `lint` / `build` clean in `docs/`.

Nine new tests: link counting from anchors, self-linking anchors not double-counted, plain-text URLs still counted, non-email steps skipped, nothing-to-score, worst-step position reporting, attachment weighting, and floor clamping. Plus a live test that `kind` survives the `GetSequencesByCampaignID` round trip. The full `internal/repository` live suite passes against dev Postgres, including `TestLiveEveryQueryPrepares` since this changes a repository query.

## Deliberately not changed

- `countTriggerTerms` double-counts overlaps: "risk free" scores both `free` and `risk free`. This also feeds `Check`, the hard gate on warmup sends, so retuning it is a product decision rather than a cleanup.
- `GetCampaignAdvancedSettings` returns defaults merged with the stored overrides, so any campaign with an advanced-settings row overrides all org settings with defaults. Pre-existing, affects every preflight setting equally, and reachable only through `PATCH /campaigns/:id/advanced`, which no web UI calls. Worth its own issue.
- `image_heavy` fires on any body under 200 characters that has an image. Plausible for short cold copy, though signature images are added after scoring so templates rarely trip it.


meszmate:
All four findings were real and are fixed in ec7b7bbe.

**Mixed links undercounted (P1).** Correct, and worse than the comment says: three labeled anchors plus three written-out URLs scored a clean 100/100. `max()` was the wrong operator. `countLinks` now counts every anchor plus any bare URL that is not already an anchor's destination, matching on the URL with trailing sentence punctuation trimmed. That keeps the self-linking-anchor case at one and the mixed case at six. Two tests added for both.

**Cancelled scoring leaves the spinner stuck (P1).** Correct: the early return did not clear `pending`, and the in-flight request's `finally` is guarded by `cancelled`, so nothing could turn it off. Applied the suggestion.

**Attachment errors produce false passes (P1).** Correct, and it contradicted the doctrine in this function's own doc comment: a check that did not run must never look like one that succeeded. A read failure now returns the same failed result the step-list read failure does.

**Verbose comments (P2).** Fair against the repo's one-line convention. Condensed the inline comments and trimmed the new doc comments.

`gofmt` and `make lint` clean, Go tests pass, `internal/repository` live suite passes against dev Postgres, `pnpm typecheck` and `lint` clean in `web/`.

Issue #187, opened 2026-08-25
[Contacts/Campaigns] Adding a contact to another campaign does not add the contact


code-withdarshan:
When selecting a contact from All Contacts and choosing Add to Campaign, the selected contact is not actually added to the chosen campaign.

Expected: The contact should be associated with the selected campaign, appear in the campaign's contact list, and update the campaign contact count.

Actual: The action does not add the contact to the campaign, even though the UI appears to accept the operation.

code-withdarshan:
It add the contact but in the campaign lead it is not seen in list i have to search then it seen in search

PR #257, opened 2026-08-29
Recipient-level engagement in the campaign Leads view (issue #250)

### LINKED ISSUE #250: Add Recipient-Level Email Engagement Tracking to Campaigns for Opens, Clicks, and Replies
One important feature is missing from Campaigns: users should be able to see which recipients opened the email, clicked on the email, and replied.

This would help users track recipient engagement and easily identify:

Who opened the email
Who did not open the email
Who clicked on the email
Who replied to the email

@meszmate

meszmate:
Closes #250

## What changes

On a campaign's **Leads** tab you can now see and filter by who opened, who did not open, who clicked and who replied.

### Server
- `POST /contacts/search` gains an `engagement` filter (`opened`, `not_opened`, `clicked`, `not_clicked`, `replied`, `not_replied`, `bounced`), built the same way as `lead_status` (`leadEngagementClause`, parameterised, scoped to the single campaign placeholder). It composes with `lead_status` as AND.
- An open is a **human** open: `opened_at IS NOT NULL AND NOT opened_machine`, the same split analytics reports as machine opens. The per-lead `campaign_lead.opened` count now follows that definition and a new `machine_opened` count carries the automated ones, so the column and the filter never disagree.
- The `not_*` values require at least one sent step, so a queued lead is neither opened nor not opened.
- Unknown values and the wrong campaign cardinality are 400s with stable codes: `invalid_lead_status`, `invalid_engagement`, `lead_filter_requires_campaign` (the existing `lead_status` validation moves onto the same codes).
- `lead_counts` gains `contacted`, `opened`, `clicked`, `replied_any` so the chips carry campaign-wide totals that match the filters.
- Export: four new columns `lead_status`, `lead_opened`, `lead_clicked`, `lead_replied`, filled when the filters name exactly one campaign. `scope=selected` now honours `filters` on top of `contact_ids` so a selected-rows export from a campaign gets those columns too.

### Web
- Leads-only **Opened / Clicked / Replied** columns (count of engaged steps, dash for sent-but-not, blank for not yet sent, "auto" for a machine-only open). The plain Contacts table is unchanged.
- The status chips above the list are now buttons that set `lead_status` on the request; a second group of engagement chips sets `engagement`. Counts and pagination come from the server for that scope, and the strip stays mounted on an empty scope so the filter can be cleared. Matching **Lead status** and **Engagement** sections in the Filters sheet.
- Export dialog offers the four campaign columns when opened from a campaign.
- Realtime: `EMAIL_OPENED` / `EMAIL_CLICKED` / `EMAIL_REPLIED` / `EMAIL_BOUNCED` are already org-scoped and already invalidate `['contacts']`, which covers the `['contacts','list',…]` leads query, so no new emit or invalidation was needed and no `refetchInterval` was added.

### Tests
`TestLiveContactEngagementFilters` (`internal/repository/contact_engagement_live_test.go`) seeds one lead per shape (human open, machine open, click, reply, bounce, sent-only, never sent) and asserts every filter value, that a machine open is not an open, that `not_opened` returns sent-but-unopened leads and excludes the never-sent one, that `lead_status` + `engagement` AND together, and that the new `lead_counts` totals match. Passes against the dev Postgres; `TestLiveEveryQueryPrepares` also passes.

### Docs
`guides/campaigns.mdx` (new "Who opened, clicked and replied" section), `api/reference/contacts.mdx` (request field, `campaign_lead`, `lead_counts`, export fields), `api/error-codes.mdx` (the three codes). No route or scope changes, so `endpoints.mdx` is untouched.

No migration, no new tables.

## Verification
- `gofmt -l ./cmd ./internal` prints nothing; `make lint` (golangci-lint + check-migrations) passes
- `web/`: `pnpm typecheck` and `pnpm lint` (0 errors)
- `docs/`: `pnpm types:check`, `pnpm lint`, `pnpm build`

## Also fixed on the way: contact export returned zero rows

While smoke-testing the export columns against the running backend, the CSV came back as a header and nothing else, for every scope. `ExportContacts` passed the **user id** into `Export`, and `ExportAll` handed that to `Search`, whose first parameter has been the **organization id** since the issue #187 org-scoping fix (Aug 25). Every contact export on `main` since then has been empty. The handler now passes the org id and the service/repository parameters are named for what they are.

## Verified against the running stack
Logged in through the API and exercised `POST /contacts/search` on a seeded campaign (24 leads: human opens, machine-only opens, clicks, replies, bounces, sent-only, queued): every `engagement` value returns the expected rows and `pagination.total`, the totals reconcile (opened 12 + not opened 9 = contacted 21, same for clicks and replies), `lead_status` + `engagement` AND together, unknown values return `invalid_engagement` and a missing campaign returns `lead_filter_requires_campaign`. Export: 8 rows for the clicked scope with the four `lead_*` columns filled, 30 rows for `scope=all`, and `scope=selected` fills the lead columns when the campaign is passed in `filters` and leaves them blank otherwise.


meszmate:
image

PR #350, opened 2026-09-07
feat(worker): sync every kind of IMAP mailbox

### LINKED ISSUE #345: IMAP: a server that drops the socket without a FIN hangs the sync pass forever, and a down server writes a warning every minute
## Problem

Follow-up to #342, which re-dials once go-imap has parked a dropped session in the Logout state. Two gaps remain in the same loop.

**Half-dead connections.** go-imap only moves to Logout when the read side sees EOF. When a NAT or firewall drops the TCP session without a FIN, the state stays authenticated and the next command blocks forever: go-imap v2 has no per-command timeout and `internal/client/smtpimap/imap` sets no deadline on the connection. The mailbox is the same zombie #342 set out to kill, it just needs a middlebox instead of a clean close to get there.

**Warning per pass.** With #342, a transport error returns `ErrMailServerUnreachable`, which `CaptureError` relays as `EMAIL_SERVER_ERROR`. The consumer writes an `email_account_errors` row and publishes a user-visible warning for each one. `nextSyncDelay` in `internal/app/worker/wmail/start_sync.go` only backs off for a provider 429, so a mail server that is down for a day produces about 1,440 rows and 1,440 dashboard warnings for that mailbox.

## Fix

- Put an idle deadline on the IMAP connection while a command is in flight (no bytes for a bounded time means the connection is dead), so a hung command fails, go-imap closes the session and the next pass re-dials through `ensureConnected`.
- Back off after consecutive transport failures: fast retry after a single drop, growing to the existing five-minute ceiling while the server stays down, and reset on the first clean pass.

### LINKED ISSUE #346: IMAP: servers without CONDSTORE cannot be connected at all (Outlook.com, Microsoft 365 over IMAP, Yahoo, many hosted servers)
## Problem

`Connect` in `internal/client/smtpimap/imap/client.go` refuses any server that does not advertise CONDSTORE with the critical `ErrMailCondStoreNotSupported`, and the whole incremental sync in `internal/app/worker/wmail/sync_imap.go` is keyed on `HIGHESTMODSEQ`.

Outlook.com and Microsoft 365 over IMAP do not advertise CONDSTORE. Yahoo does not either (it has a proprietary `XYMHIGHESTMODSEQ`). Plenty of hosted cPanel and older Exchange servers do not. Today those mailboxes fail at load with a critical error and never sync, even though the connect dialog accepted the credentials.

## Fix

Keep CONDSTORE as the preferred path and add an RFC 3501 baseline for everything else, chosen per connection from the capabilities:

- track `UIDNEXT` per folder next to `HIGHESTMODSEQ` (new column on `unibox_mailboxes`, new field on `models.Mailbox`), request it in the folder listing and hand it back on load like the mod-sequence
- without CONDSTORE, new mail in a folder is `UID SEARCH UID <saved uidnext>:*` and the cursor advances the same way the mod-sequence does (held while a message is deferred)
- read state and flags are mirrored by a bounded periodic flags scan over the newest window of each folder, diffed against the previous scan, with only the changed UIDs fetched
- `SELECT` only asks for CONDSTORE when the server has it

The backfill already uses `SEARCH SINCE` and needs no change. The fake server tests in `sync_imap_test.go` should cover both paths, and the client should get tests against go-imap's in-process `imapmemserver`, which has no CONDSTORE and so is exactly the server this is for.

### LINKED ISSUE #347: IMAP: a server without LIST-STATUS looks like an empty account with no error
## Problem

`Folders` in `internal/client/smtpimap/imap/client.go` asks `LIST` to return `STATUS` (RFC 5819 LIST-STATUS) and skips any folder whose status came back nil. On a server without the extension every folder is skipped, the sync pass sees zero folders, retires every saved cursor through the deletion sweep, and reports a clean pass. Nothing is logged and nothing is stored, ever.

## Fix

When the server does not advertise LIST-STATUS, run a plain `LIST`, then one `STATUS` per selectable folder for `UIDVALIDITY`, `HIGHESTMODSEQ` (when CONDSTORE is present) and `UIDNEXT`. Keep the single round trip when the extension is there. A folder whose STATUS fails should be skipped with a log line, not silently.

### LINKED ISSUE #348: IMAP: the 30-folder cap fails the whole mailbox silently, and #342 makes it count nested folders
## Problem

`config.MaxEmailFolders` is 30. When a listing exceeds it, `Folders` returns `ErrMailFoldersMax` and the pass ends. That error code is not in `mailErrorToJobEventType`, so it is never relayed: no `email_account_errors` row, no dashboard warning, only a worker log line. The mailbox looks connected and never syncs.

#342 switched the listing to `LIST "" "*"` (needed so Gmail's `[Gmail]/Sent Mail` and Dovecot's `INBOX.*` are reached), which means nested folders now count. On Gmail the seven `[Gmail]/` folders count too, so a Workspace account with around 22 labels hits the cap. A long-lived personal mailbox on any server hits it easily.

## Fix

- count only selectable, non-virtual folders
- raise the cap (the instance limits page under Instance settings shows the value and needs nothing else)
- when the listing still exceeds it, keep INBOX and the special-use folders first, take the first N of the rest, and sync those rather than failing the mailbox
- relay the overflow once per worker session as a warning so the user sees which folders are not followed

### LINKED ISSUE #349: IMAP: folders on servers without SPECIAL-USE, or with localized names, all file as inbox
## Problem

`imapCanonicalFolder` in `internal/app/worker/wmail/sync_imap.go` trusts RFC 6154 special-use attributes and falls back to a short list of English leaf names. Older Exchange, cPanel and many hosted Dovecots attach no attributes, and their Sent, Drafts, Junk and Trash folders are localized ("Enviados", "Gesendet", "Éléments envoyés", "Elküldött elemek"). Everything unrecognised files as inbox, so the unibox shows the customer's sent and junk mail in the inbox scope and `AppendToSent` finds no Sent folder to file copies in.

The leaf is also computed by guessing between `/` and `.` as the delimiter, while `LIST` reports the real one per folder.

## Fix

- one shared table of folder names, in the common languages Warmbly customers use, feeding `imapCanonicalFolder`, `sentMailbox` and the spam guard in `warmup_actions.go`
- carry the delimiter `LIST` reports on each folder and split the leaf on it, keeping the `/` and `.` guess only for saved rows that predate the field
- tests for each language and for prefixed names (`INBOX.Sent`, `INBOX/Gesendet`)

meszmate:
Closes #345, closes #346, closes #347, closes #348, closes #349.

Follow-up to #342, which fixed the dropped-session zombie and Gmail's nested folders. That left the IMAP client working well on Gmail and silently broken, or refusing to connect at all, on several other providers. This makes a plain RFC 3501 server a first-class mailbox.

## What was broken

**Servers without CONDSTORE could not be connected at all** (#346). `Connect` refused them with a critical error and the whole incremental sync was keyed on `HIGHESTMODSEQ`. Outlook.com, Microsoft 365 over IMAP and Yahoo do not advertise it, nor do many hosted cPanel and older Exchange servers. The connect dialog accepted the credentials and the mailbox then failed at load.

**Servers without LIST-STATUS looked like empty accounts** (#347). Every folder whose status came back nil was skipped, so the pass saw zero folders, retired every saved cursor through the deletion sweep, and reported success. Nothing logged, nothing stored, forever.

**A peer that vanished without a FIN parked a command forever** (#345). go-imap bounds a response it has started reading, but clears the deadline between responses, which is exactly where the reader waits for the answer to the command just sent. A NAT or firewall dropping the mapping left the mailbox a zombie until the worker restarted, the same failure #342 fixed for cleanly-closed sessions.

**The folder cap failed the whole mailbox silently** (#348). Thirty folders, and `MAX_FOLDERS_REACHED` was not in the event-type map, so it was never relayed: no error row, no warning, one worker log line. #342 made the listing recursive, so nested folders now count and a Workspace account with ~22 labels hit it.

**Folders on servers without SPECIAL-USE all filed as inbox** (#349). The name fallback knew English only, so a mailbox whose Sent folder is "Gesendete Elemente" showed its sent mail in the inbox scope and got no sent copies filed.

## What this does

- **UIDNEXT as a second cursor.** `unibox_mailboxes` gains `uid_next` (migration 000131) and `models.Mailbox` a matching field. With CONDSTORE nothing changes. Without it, new mail is `UID SEARCH <cursor>:*`, filtered because IMAP answers `n:*` past the end with the last message, and the cursor is held on a deferred batch exactly as the mod-sequence is.
- **A periodic flag scan** mirrors read state where the server cannot report changes: one FETCH of the newest window per folder every ten minutes, diffed against the previous scan, relaying only what moved. The first scan baselines rather than relaying the window. New mail does not wait for it.
- **STATUS per folder** when LIST-STATUS is missing, with a folder the server refuses skipped and logged rather than taking the account with it.
- **An idle deadline** on the transport, filling in the stretch go-imap leaves uncovered. It never shortens a deadline go-imap set, so a slow body fetch that is making progress is not cut.
- **The cap is now 100**, counts only selectable non-virtual folders, keeps INBOX and the special folders first, and relays the overflow once per worker session as a warning.
- **Localized folder names** in a dozen languages for sent, drafts, spam, trash and archive, shared by the sync loop, the Sent resolver and the warmup spam guard. The guard was also tightened from substring to exact leaf match, so a user folder called "Spam reports" is no longer a MOVE target.
- **Transport back-off**: a widening retry to the existing five-minute ceiling while the server is down, and one warning per outage instead of one per pass.

## Verification

- `make lint` (golangci-lint plus `make check-migrations`) and `go test ./internal/...` pass; `pnpm types:check` and `pnpm lint` pass in `docs/`.
- New tests run against go-imap's in-process server configured with **no CONDSTORE and no LIST-STATUS**, which is the exact shape this package used to refuse: connect, nested listing, per-folder STATUS, the UIDNEXT search including the past-the-end case, the flag scan carrying Message-IDs, and the re-dial after a dropped socket.
- The transport wrapper has its own tests for all three behaviors: it bounds the cleared stretch, releases between commands so an idle session is not cut, and does not shorten a deadline go-imap set.
- Worker-side tests cover the UIDNEXT path, the held cursor on a deferred batch, a quiet folder costing nothing, the flag scan baselining then relaying, the scan staying periodic and skipped under CONDSTORE, and the overflow warning being sent once over three passes.

Writing that last test is what caught the folder-limit code missing from `mailErrorToJobEventType`, so the new warning would have gone nowhere.

## Not covered

Plaintext IMAP is still refused. Proton Bridge listens without TLS on localhost, so it needs an explicit opt-in rather than a silent fallback; that is worth its own issue and its own decision.

## Summary by CodeRabbit

- **New Features**
  - Improved IMAP synchronization for servers without CONDSTORE, including incremental message discovery and periodic flag updates.
  - Added support for nested folders, localized folder names, Gmail labels, and additional IMAP providers.
  - Increased the synchronized folder limit from 30 to 100, prioritizing important folders.

- **Bug Fixes**
  - Improved recovery during mail-server outages and clearer folder-limit warnings.
  - Improved handling of duplicate folder identifiers and localized system folders.

- **Documentation**
  - Added troubleshooting guidance for skipped IMAP folders and synchronization behavior.


meszmate:
Thanks, all three were real. Fixed in the latest commit.

**NIL delimiter.** Correct, and it was inconsistent: `Folders` guarded `f.Delim != 0` but `sentMailbox` did not, so a server reporting no hierarchy got `"\x00"` as its separator, `leafWithDelim` matched nothing, and `INBOX.Sent` stopped resolving as Sent. There is now one `delimString` helper both call, with a test for the NIL case.

**Deadline race.** Also real. The sync loop holds only the lifecycle read lock, so a fetch and a warmup STORE do overlap on one session, and the atomic counter left a window where the last release cleared the deadline just after another command armed it. The counter and every deadline call now sit under one mutex, including the write side. `TestIdleConnStaysArmedWhileAnyCommandIsInFlight` runs fifty commands starting and finishing under a long-running one and then checks the outer command is still bounded; the package passes under `-race`.

**Delimiter in classification.** Fixed by threading it rather than by adding a parameter: `CanonicalFolder`, `BackfillEligible` and `IsVirtualFolder` now take the `models.Mailbox`, which already carries `Delim` from the listing, so the worker wrappers pass it for free. `TestCanonicalFolderUsesTheServerDelimiter` covers `|` and `\` as you suggested, plus the no-delimiter fallback and a folder whose name legitimately contains a dot on a `/` server.

`IsVirtualFolder` deliberately still hard-codes `/` for its Gmail namespace check, since that is Gmail's delimiter specifically, not the connected server's.

meszmate:
Good catch on the inconsistency, though I have kept the sorting and fixed the test instead.

The ranking is deliberate on every listing, not only when something is cut. The sync pass walks folders in the order `Folders` returns and can run out of sync budget partway through, so walking the inbox and the special folders first means a reply to the customer's own outreach lands before a mailing list in a user folder does, whatever order the server happened to list them in. Returning the server's order early would give that up for every mailbox under the cap, which is almost all of them.

What was actually wrong is what you spotted underneath it: `TestRankFoldersKeepsEverythingUnderTheCap` carried a comment saying the cap "must not reorder a listing that fits" while asserting nothing about order, so it documented the opposite of the behaviour. The test now asserts the ranked order explicitly, and `rankFolders` says why the ranking is unconditional.

meszmate:
Fixed: the UIDVALIDITY conflict now has its own `FOLDER_ID_CONFLICT` code, its own entry in the worker's event mapping, and its own title and guidance pointing at renaming the folder rather than deleting folders to get under a cap. The shared code meant the drawer told a user to delete folders for a problem where neither the cap nor deleting anything was involved.

Added `TestFolderProblemsReachTheUserWithTheirOwnCodes`, which checks both reasons arrive as distinct codes and as warnings rather than anything that would deactivate the mailbox. That is the same class of gap as the earlier one where the folder-limit code was missing from the mapping entirely and the warning went nowhere.

PR #354, opened 2026-09-07
Read the dashboard's Sentry DSN from runtime config, so a self-host reports nowhere by default

meszmate:
Closes #351.

`web/src/main.tsx` initialised `@sentry/react` with a literal DSN pointing at Warmbly's own Sentry organisation, with `sendDefaultPii: true` and no way to override or disable it. The dashboard image is the same for hosted and self-host, so every self-hosted install was reporting its users' browser errors, URLs and IPs to Warmbly.

## What changed

- `web/src/lib/observability.ts` is the one place the SDK is touched. `initErrorReporting()` returns without calling `Sentry.init` when no DSN is configured, and `captureException()` is a no-op in that case.
- The DSN comes from runtime config: `runtimeEnv("SENTRY_DSN", import.meta.env.VITE_SENTRY_DSN)` in `web/src/lib/information.ts`, injected by `web/docker-entrypoint.sh` from `WARMBLY_SENTRY_DSN`, matching the backend's "empty is unset" rule.
- The literal DSN is gone from the repo. **The exposed key still needs rotating in Sentry** — that is the one part of the issue this PR cannot do.
- `web/src/app/auth/login/page.tsx` calls the wrapper instead of the SDK directly.
- `docker-compose.yml` passes `WARMBLY_SENTRY_DSN` through to the web container, defaulting to empty.
- `configuration.mdx` documents the variable; `data-control.mdx` gains an "Error reporting" section stating that an instance reports errors nowhere unless the operator points it somewhere.

The SDK is imported statically rather than lazily on purpose: its global handlers need to be installed before the first render, since a broken deploy fails during boot and a chunk still in flight would miss exactly that error. Not initialising it costs a self-hoster some dead bundle weight and zero network calls.

`admin/` still has no Sentry; when it gets one (#353) it follows the same rule.

## Verification

- `pnpm typecheck` and `pnpm lint` in `web/`: clean, no new warnings from the changed files.
- `pnpm types:check` and `pnpm lint` in `docs/`: clean.
- `grep` for the literal DSN and org id across the repo: no hits.

## Summary by CodeRabbit

- **New Features**
  - Added optional browser error reporting for the dashboard when a Sentry configuration is provided.
  - Handled application and passkey initialization errors can now be captured through the centralized reporting system.
  - Error reporting remains disabled by default when no configuration is supplied.

- **Documentation**
  - Documented error-reporting configuration across supported services.
  - Clarified that reporting is configured independently for each process and occurs only when enabled.
[...]

Issue #351, opened 2026-09-07
Dashboard hardcodes Warmbly's Sentry DSN, so self-hosted installs send errors (with PII) to Warmbly's Sentry


meszmate:
## What happens

`web/src/main.tsx` initialises `@sentry/react` with a literal DSN pointing at Warmbly's own Sentry organisation (`...ingest.us.sentry.io/...`), with `sendDefaultPii: true` and no way to override or disable it. `web/src/app/auth/login/page.tsx` also calls `Sentry.captureException` directly.

The dashboard image is the same for hosted and self-host, so every self-hosted Warmbly dashboard reports its users' browser errors, URLs and IPs to Warmbly's Sentry. That contradicts the self-host promise (nothing phones home) and the backend's own stance in `internal/observability/sentry.go`, where a DSN is "the operator's choice, not a requirement of the software".

## Expected

- The web DSN comes from runtime config like the other web settings: `runtimeEnv("SENTRY_DSN", import.meta.env.VITE_SENTRY_DSN)` in `web/src/lib/information.ts`, injected by `web/docker-entrypoint.sh` from `WARMBLY_SENTRY_DSN`.
- No DSN means Sentry is not initialised at all (no script, no network calls), matching the backend's "empty is unset" rule documented in `docs/content/docs/development/configuration.mdx`.
- The literal DSN is removed from the repo and the exposed key rotated in Sentry.
- `configuration.mdx` documents the new variable for the web container, and `data-control.mdx` states that a self-hosted instance reports errors nowhere unless the operator sets it.

## Notes

- `admin/` has the same runtime config mechanism (`admin/src/lib/runtimeConfig.ts`, `admin/docker-entrypoint.sh`) and no Sentry today; when it gets one it must follow the same rule.
- Part of the wider Sentry coverage work tracked separately.

PR #358, opened 2026-09-07
Complete Sentry coverage across every service, optional everywhere, behind one Go wrapper

### LINKED ISSUE #353: Complete Sentry error reporting across every service, optional everywhere, with releases and source maps, and a single Go wrapper
## Goal

One error reporting story that covers every runtime in the repo, is fully optional (no DSN means nothing is sent, in every environment), and can be pointed at Sentry Cloud, a self-hosted Sentry or a Sentry-compatible server such as GlitchTip by setting a single variable per service. The Sentry SDKs stay the integration layer because they are the only first-party set covering Go, Elixir, Rust, React and Astro; the server behind them is the operator's choice.

Depends on #351 (the dashboard's hardcoded DSN), which should land first.

## Current state (verified)

- Go: `internal/observability/sentry.go` initialises Sentry for `backend`, `consumer` and `worker` only. `cmd/forms` has no init. 89 Go files import `sentry-go` directly and call `sentry.CaptureException` etc.
- Web: `@sentry/react` with a literal DSN (#351). No release tag, no source map upload, so production stack traces are minified.
- Admin: no error reporting.
- Realtime (Elixir): `:sentry` configured in `realtime/config/runtime.exs` from `SENTRY_DSN`, empty treated as unset. Good; keep as the reference behaviour.
- Tracking (Rust): no `sentry` crate, no error reporting at all.
- Forms app (`forms/`, public form pages): none.
- Docs: `configuration.mdx` lists `SENTRY_DSN` as optional per service.

## Scope

1. **Go wrapper.** Add a thin package (for example `internal/observability/errs` or extend `observability`) exposing `CaptureException(ctx, err, tags...)`, `CaptureMessage`, `Flush`, and a request-scoped hub helper. Replace direct `sentry.` calls across the 89 files with it, so the SDK is imported in one place. Keep `sentry.Recover` style middleware behaviour intact. This is mechanical; do it in one PR with no behaviour change.
2. **Every Go service initialised.** `cmd/forms` gets `InitSentry`. `cmd/warmblyctl`, `cmd/updater`, `cmd/migrate`, `cmd/seed`, `cmd/sandbox` stay without (operator tools, local output is enough). `cmd/cli` never reports anywhere: it runs on customers' machines.
3. **Release and environment tags.** Every service tags events with the release (git tag or short SHA, injected via `-ldflags -X` in the Makefile and `scripts/build-*.sh`, and read by `InitSentry`) and environment. The web and admin builds set `release` from the same value at build time.
4. **Web and admin.** DSN from runtime config (after #351), `admin/` gets the same optional `@sentry/react` init through `admin/src/lib/runtimeConfig.ts` and `admin/docker-entrypoint.sh`. Source maps uploaded from CI only when a Sentry auth token is present (Sentry Vite plugin), never as a required step, so forks and self-host builds do not need a Sentry account.
5. **Forms app (`forms/`).** Optional `@sentry/browser` init from the same runtime config mechanism the Go forms server injects; public pages must stay light, so load it only when a DSN is present.
6. **Tracking (Rust).** Add the `sentry` crate behind `SENTRY_DSN`; a panic or a backend circuit-breaker trip should reach the same place as backend errors. No DSN means the layer is not installed.
7. **Realtime.** Add release tagging to match; already optional.
8. **Docs.** `configuration.mdx`: one `SENTRY_DSN` row per service including web, admin, forms and tracking, plus the release variable; a short paragraph in `docs/content/docs/development/instance-health.mdx` or `troubleshooting.mdx` on pointing every service at Sentry Cloud (free Developer plan: 5k errors a month) or a Sentry-compatible self-hosted server. `data-control.mdx`: a self-hosted instance reports errors nowhere unless the operator sets a DSN.

## Acceptance

- `grep -rl getsentry/sentry-go cmd internal` returns only the wrapper package (plus the middleware that needs the hub).
- Starting every service with no `SENTRY_DSN` produces no outbound connection to any Sentry host; starting them with one produces events tagged with `release`, `environment` and the service name.
- A thrown error in the production dashboard shows a readable stack trace in Sentry (source maps applied).
- `make fmt`, `make lint`; `pnpm typecheck` and `pnpm lint` in `web/`, `admin/`, `forms/`; `cargo check` in `tracking/`.

## Out of scope

- Switching to PostHog error tracking. It lists Go, Elixir, Rust and React, but it is young and would mean replacing every SDK; the wrapper in step 1 makes that a one-file change later if wanted.
- Performance tracing and profiling. Can be turned on later through the same wrapper.

meszmate:
Closes #353. Stacked on #354 (the hardcoded dashboard DSN), which it depends on — **merge #354 first**, then this retargets to `main` automatically.

## 1. One Go wrapper

`internal/observability/errs` is now the only package in the repo that imports `getsentry/sentry-go`:

```
$ grep -rl getsentry/sentry-go cmd internal
internal/observability/errs/errs.go
```

549 call sites across 88 files were rewritten mechanically (`sentry.CaptureException` → `errs.CaptureException`, and so on). The wrapper exposes `CaptureException`, `CaptureMessage`, `Recover`, `Flush`, `Option` builders (`errs.Tag`, `errs.Extra`), and context-aware variants plus `errs.Hub(ctx)` / `errs.NewContext(ctx)` for the request-scoped hub the issue asks for. `observability.InitSentry` keeps its signature and now resolves the DSN and the release and hands them to `errs.Init`.

Three sites used `sentry.WithScope` and became `Option` calls. One of them, `internal/infrastructure/db/error.go`, captured the exception *before* opening the scope, so every database error has been reported with none of its tags; the tags now actually apply. That is the only behaviour change in the migration.

## 2. Every service initialised, and tagged

- `cmd/forms` gets `observability.InitSentryEnv("forms")` — an env-only variant, because the forms service has no database and no secrets backend and should not build a `config.Config` just to read one variable.
- `cmd/warmblyctl`, `cmd/updater`, `cmd/migrate`, `cmd/seed`, `cmd/sandbox` stay without, and `cmd/cli` reports nowhere: it runs on customers' machines.
- Release and environment tags needed no build plumbing. Every Go image already stamps `internal/version` via `-ldflags`, so `observability.Release()` reads it. Tracking and realtime read `WARMBLY_RELEASE`, which their Dockerfiles now set from the same `VERSION`/`COMMIT` build args the Go images take, passed by both workflows.

## 3. Web, admin, forms, tracking, realtime

- **admin** gets the same optional `@sentry/react` init as the dashboard, through `admin/src/lib/runtimeConfig.ts` and `admin/docker-entrypoint.sh`. No DSN, no init.
- **Source maps** upload from the Vite plugin only when `SENTRY_AUTH_TOKEN` (a build secret), `SENTRY_ORG` and `SENTRY_PROJECT` are all present. A fork or a self-host build sets none of them: the plugin is not in the plugin list, `build.sourcemap` stays off, and the image is what it was. Source maps are deleted after upload so they never ship.
- **forms/** loads `@sentry/browser` as its own chunk and only when the Go shell stamped a DSN into the page. Verified against a real build: `index.html` references only the 388 kB app chunk, and the SDK sits in a separate chunk nothing fetches without a DSN. Public pages stay light, and `sendDefaultPii` is off — a form page carries a stranger's answers.
- **tracking (Rust)** gets the `sentry` crate (rustls, no OpenSSL) behind `SENTRY_DSN`, wired into the existing `report_issue`/`report_error` functions every failure path already goes through, plus the panic hook. An invalid DSN logs and disables reporting rather than stopping the service: the pixel is the product.
- **realtime** now tags `release` and takes its environment from `APP_ENV` like everything else.

## 4. Also fixed: `.gitignore` was swallowing `forms/`

`/forms` was meant to ignore the compiled `cmd/forms` binary, but it matches the `forms/` source tree too, so **every new file added under `forms/` was silently invisible to git**. The binary cannot land at the repo root anyway (`go build -o forms` refuses to overwrite a directory), so the entry is removed. `forms/.gitignore` already covers `dist/` and `node_modules/`.

## Verification

- `grep -rl getsentry/sentry-go cmd internal` → the wrapper only.
- `gofmt -l cmd internal` → empty. `make lint` (incl. `make check-migrations`) → passes.
- `pnpm typecheck` + `pnpm lint` in `web/`, `admin/`, `forms/` → 0 errors, no new warnings.
- `cargo check`, `cargo clippy --all-targets` (0 warnings), `cargo fmt --check` in `tracking/`.
- `go test ./internal/formserver/` passes, including a new test that the browser DSN is stamped, that an empty one leaves the tag empty, and that an operator-supplied value cannot close the tag and inject markup into a public page.
- `docs/` `pnpm types:check` + `pnpm lint` clean.

## Out of scope, as the issue says

Performance tracing and profiling, and switching to PostHog error tracking — the wrapper makes the latter a one-file change if it is ever wanted.


meszmate:
Superseded by #367. GitHub auto-closed this when its base branch (the #351 fix, merged as #354) was deleted, and a closed PR cannot be retargeted — #367 is the same branch reopened against `main`.

PR #364, opened 2026-09-07
Cookieless PostHog analytics for the hosted properties, with first-party signup attribution

### LINKED ISSUE #352: Cookieless PostHog analytics for the hosted properties (warmbly.com and the dashboard) with first-party signup attribution, never loaded on self-host
## Goal

Understand who visits warmbly.com, which pages and channels bring signups, and what new customers do in the dashboard (funnels such as signed up, connected a mailbox, launched a campaign), with **no cookie banner**: nothing stored in the visitor's browser and no consent prompt.

Research with primary sources is in `resources/cookieless-analytics-research.md` (added in this branch of work; see the comparison table). Decision: **PostHog Cloud EU in cookieless mode** for aggregate analytics and product funnels, plus a **first-party attribution record at signup** stored in Warmbly's own database for anything that has to survive longer than a day.

## Why PostHog cookieless

- `cookieless_mode: 'always'` stores nothing client-side; the server derives the visitor from a daily-rotated salt plus IP, root domain and user agent, and the salt is deleted after the day (PostHog docs and `cookieless-manager.ts` in the PostHog repo).
- The hash uses the registrable root domain, so warmbly.com and app.warmbly.com are one visitor and one session within a day with no cross-domain wiring.
- EU Cloud (Frankfurt), MIT core, 1M events a month free.
- Limits to respect: do not call `identify`, set `person_profiles: 'never'`; `alias` is dropped at ingestion. No persistent user profiles, and no per-person retention over weeks. That is what the first-party record below is for.

### 1. Marketing site (`site/`)

- Load the PostHog snippet from `site/src/layouts/Layout.astro` only when a `PUBLIC_POSTHOG_KEY` (and `PUBLIC_POSTHOG_HOST`, default EU) build variable is set. Unset means no script tag at all. Document the variables where the site build is documented.
- Cookieless config: `cookieless_mode: 'always'`, `person_profiles: 'never'`, `autocapture: false`, `capture_pageview: true`, `disable_session_recording: true`, `respect_dnt: true`.
- CTA links to the dashboard signup carry the landing path and any UTM parameters through to `/signup` on the dashboard origin (query string, not storage).

### 2. Dashboard (`web/`)

- Same snippet, gated at runtime: `runtimeEnv("POSTHOG_KEY", ...)` in `web/src/lib/information.ts`, injected by `web/docker-entrypoint.sh` from `WARMBLY_POSTHOG_KEY` / `WARMBLY_POSTHOG_HOST`. Unset means PostHog is never initialised. Self-host images therefore contain the code path but never load it; the installer must not ask about it and `.env` templates must not mention it as something to fill in.
- Same cookieless config as the site. Session replay stays off: it would record mailbox and contact screens.
- Product events, all without user or org ids in properties: `signup_completed`, `mailbox_connected` (provider), `campaign_launched`, `trial_started`, `subscription_started` (plan). Prefer firing the money events server-side (below) so they are exact.
- The `admin/` panel gets nothing. It is the operator surface, only staff use it, and it ships to self-hosters too. Documented as out of scope so nobody adds it later.

### 3. Server-side events (backend)

- A small `internal/observability/analytics` (or similar) client that posts to the PostHog capture API when `POSTHOG_KEY` is set on the backend, forwarding the originating request's `User-Agent` and client IP so PostHog's cookieless hash matches the browser session. Off when unset, which is the self-host default.
- Fire `signup_completed` from registration and `subscription_started` (with plan and amount) from the Stripe webhook handler.

### 4. First-party attribution (the part that does not depend on PostHog)

- Migration (next number after the highest on `main`, and `make check-migrations`): nullable acquisition columns on `organizations` or a small `organization_acquisition` table: `landing_path`, `referrer_host`, `utm_source`, `utm_medium`, `utm_campaign`, `utm_term`, `utm_content`, captured once at signup and never updated.
- The registration flow reads them from the signup URL query string and stores them. Nothing stored before signup, so still no client-side storage.
- Register the table (or note the columns) in `internal/app/orgtransfer/spec.go` so the data travels with an org export, per the repo rule that no org-scoped table is done until it is in the transfer spec.
- Surface: a column and filter in the admin panel org list, so revenue by channel is a SQL join against subscriptions rather than a vendor report.

### 5. Docs and policy

- `docs/content/docs/development/configuration.mdx`: the new variables, marked as hosted-only, unset by default.
- `data-control.mdx`: state that a self-hosted instance loads no analytics and sends no usage data.
- Privacy policy on the site: name PostHog (EU) as a processor for aggregate analytics without cookies, and the acquisition fields captured at signup.
- Note the legal position from the research: cookieless pageview analytics has a strong no-banner footing (CNIL 2025 grid accepts salted, daily-rotated hashes), but the CNIL exemption excludes acquisition-channel and conversion measurement. Recording channel at signup as part of the account record is a product decision made here deliberately; keep it minimal and first-party.

## Acceptance

- A fresh self-host install (`install.sh`, default answers) produces a dashboard and site build with zero requests to any PostHog host. Verify by grepping the built assets for the host and by checking the runtime config output.
- On warmbly.com, DevTools shows no cookies and no localStorage entries written by the snippet.
- A signup started from a UTM-tagged link ends with the UTM fields on the organization, and a `signup_completed` event in PostHog within the same day's session.
- `pnpm typecheck` and `pnpm lint` in `site/` and `web/`; `make fmt`, `make lint`, `make check-migrations`; `make installer-check` if the installer or `.env` template changes.

## Out of scope

- Session replay, heatmaps, surveys, feature flags: each stores or records more than a cookieless pageview and needs its own decision.
- Replacing Sentry with PostHog error tracking (see the Sentry coverage issue).
- Warmbly's own customer-facing website tracking feature is unrelated: it identifies known contacts and keeps a visitor id, which is why it has a consent mode.

meszmate:
Closes #352. Now targets `main` — the Sentry work it was stacked on landed as #367.

Research, checked against primary sources and cited, is in `resources/cookieless-analytics-research.md`. The issue referenced that file but it had never been written; everything below was verified against PostHog's docs and its actual ingestion source rather than from memory.

## No cookie banner, and no self-host telemetry

- **Cookieless.** `cookieless_mode: 'always'` stores nothing in the browser. PostHog derives the visitor server-side from a daily-rotated salt plus IP, root domain and user agent, and deletes the salt at the end of the day. That only holds if nobody is identified, so `person_profiles: 'never'` is set everywhere and no event property carries a user id, an organization id or an email.
- **Hosted-only.** Every key is unset by default. Verified against real builds:
  - `site/`: a build with no `PUBLIC_POSTHOG_KEY` has **zero** occurrences of `posthog` in `dist/` — no script tag is emitted at all.
  - `web/`: the 276 kB SDK is a separate chunk that `index.html` does not reference. Without a key the dynamic import never fires and no PostHog host is contacted. The code path ships (one image serves hosted and self-host, as the issue specifies); the request does not.

The installer asks nothing about it and no `.env` template mentions it.

## What was built

**Marketing site.** Snippet in `Layout.astro`, gated on the build variable. A second inline script carries the landing path and any UTM parameters onto every `/register` link — one place, because there are 39 such links across 20 files and a new CTA should not have to remember.

**Dashboard.** Runtime-gated via `WARMBLY_POSTHOG_KEY`. `autocapture: false` deliberately: the dashboard is behind a login and autocapture would ship contact names and subject lines in element text. Session replay off — it would record mailbox and contact screens. `capture_pageview: 'history_change'`, since plain `true` reports one pageview per session in an SPA. A `before_send` hook masks record uuids out of `$current_url` / `$pathname` / `$referrer`, so a pageview never carries an identifier and the report reads one row per screen instead of one per record. Events are a closed TypeScript union so a typo is a build error.

**Server-side.** `internal/observability/analytics` posts to `/i/v0/e/` with `distinct_id: "$posthog_cookieless"`, `$cookieless_mode`, `$raw_user_agent`, `$ip` and `$host` — the exact contract read out of `rust/common/cookieless/src/constants.rs` and `nodejs/src/ingestion/common/cookieless/cookieless-manager.ts`. `New()` returns `nil` without a key and a nil `*Client` is a working no-op, so there is one path to test. A wrong key or an unreachable host logs once per process rather than silently forever or on every event.

`signup_completed` and `trial_started` fire from the signup flow carrying the request's identity so they join the browser's session; `subscription_started` fires from the Stripe webhook, where there is no browser request to join — noted in the code, since it is a count and a plan mix rather than the end of a funnel.

**First-party attribution.** Migration `000133` adds `organization_acquisition` (landing path, referrer host, five UTM fields), written once at signup with `ON CONFLICT DO NOTHING` — the row is the account's origin, not a mutable setting. Registered in `internal/app/orgtransfer/spec.go` under `Tables`, per the repo rule. The admin org list gets a Channel column (hidden by default; most signups are direct) and `utm_source` / `utm_medium` / tagged / direct filters, so revenue by channel is a SQL join rather than a vendor report.

Plumbed as `SignupAttribution{ReferralCode, Invite, Acquisition}`, which also takes `createAccount` from six positional parameters down to five.

**The admin panel gets no analytics**, documented as deliberate so nobody adds it later.

## Docs and policy

`configuration.mdx` gains a Product analytics section; `data-control.mdx` states plainly that a self-hosted instance loads no analytics and sends no usage data, and tells the reader how to check. The privacy policy names PostHog EU as a processor, explains the cookieless mechanism and why there is no banner, and describes the acquisition fields and why they are kept.

The research file records the legal position honestly: the CNIL's 2025 grid accepts a salted daily-rotated hash without consent, but that exemption **excludes** acquisition-channel measurement. Recording the channel at signup is a product decision made deliberately, kept minimal and first-party.

## Verification

- `make fmt`, `make lint`, `make check-migrations` (133 migrations, no duplicates — the acquisition migration was renumbered from `000132` to `000133` when `main` landed its own `000132`).
- `pnpm typecheck` + `pnpm lint` in `web/`, `admin/`, `forms/`: 0 errors, no new warnings. `pnpm build` in `site/` (its CI gate): clean, 45 pages.
- **Live tests against a real Postgres** (`WARMBLY_TEST_DB`): the acquisition round-trip and the admin list. These cover the parts a unit test cannot — the LEFT JOIN, the four columns added to the admin projection, all three channel filters, that a second signup link cannot rewrite where a workspace came from, and that the join does not fan out a row.
- **The capture client is tested against a stub PostHog**, asserting the ingestion contract (sentinel distinct id, mode flag, three hash inputs) and that no event property names a person.
- Unit tests on the acquisition normalizer caught a real bug: `url.Parse("pricing")` produced a relative path that slipped through instead of being dropped.
- **The CTA decorator was exercised in jsdom against the real built pricing page**: all six register links carry the UTM parameters, the landing path and the referrer host, and the login links are left alone. It also caught a bug — the script sits in `<head>` and originally ran before the anchors existed.

## Summary by CodeRabbit

- **New Features**
  - Admins can view organization acquisition details, including UTM data and landing paths, in an optional table column.
  - Added filters for acquisition source, medium, campaign, tagged links, and direct signups.
  - Signup links now preserve attribution data such as UTM parameters, landing paths, and referrers.
  - Added optional cookieless product analytics for signups, mailbox connections, campaigns, and subscriptions.

- **Documentation**
  - Added configuration, privacy, and data-control guidance for analytics and attribution tracking.


meszmate:
All eight fixed. Notes on the three that were more than a quick edit:

**`normalizeHost` retaining query and fragment data.** Confirmed and it was a real leak: `url.Parse("example.com?email=alice@example.com")` reads the whole thing as a relative path, so `Hostname()` is empty and the fallback ran, and the fallback only cut on `/` and `:`. The reduction now also cuts on `?`, `#` and `@`, and — more importantly — the *result* is checked against a hostname pattern rather than trusted, so anything that is not a hostname is dropped instead of stored. Regression cases added for query, fragment, userinfo and a non-host string.

Separately, and prompted by the same finding: email-shaped values are now redacted out of **every** acquisition field before they reach the database or an analytics property, since `POST /auth/register` accepts these verbatim. The comment on that regex says plainly what it is — a shape check for the one identifier worth catching, not a PII scrubber — because pretending otherwise would be worse than not having it.

**`subscription_started` idempotency.** Agreed on the problem, but an outbox is a lot of machinery for an analytics count. Instead the event now hangs off the persisted trial-to-paid transition (`wasTrialOnly && sub.HasPaidSubscription()`) rather than off the `customer.subscription.created` delivery, and fires after the write. A redelivered webhook finds the subscription already carrying a Stripe id, so the transition no longer reads as new and nothing is re-emitted. That makes it exactly as idempotent as the premium-worker migration sitting two lines below it, which relies on the same guard — and it is also more correct semantically: "started paying" rather than "Stripe sent a created event".

**The CNIL claim.** You were right and I checked the source rather than take it on trust. [CNIL's page](https://www.cnil.fr/fr/cookies-solutions-pour-les-outils-de-mesure-daudience) says nothing about hashing schemes, salt rotation or IP-plus-user-agent constructions; its conditions are a strictly-limited audience-measurement purpose, anonymous statistics only, no cross-site following, and no cross-referencing or third-party transmission — and it notes some offerings fall outside regardless of configuration. The research file now separates the two questions that were being conflated (ePrivacy Article 82 attaches to *storing or reading on the device*, which is a mechanism claim we can verify; GDPR still applies to the server-side IP and user-agent processing regardless), states the conditions as published, says the construction is *designed* against them, and says explicitly that whether a deployment qualifies is a deployment-specific assessment which that document does not make. It now opens with a "not legal advice, get an assessment" note.

That also surfaced an inconsistency in the privacy policy you did not flag: its legal-bases list said analytics runs on **consent**, which contradicts shipping without a banner. It now names legitimate interest for the aggregate cookieless measurement and the acquisition record, with the right to object, and keeps consent for marketing email.

The rest: invited signups now emit `signup_completed` (they returned early, before the count); both OAuth paths capture the validated `expected.provider`, and the Warmbly Cloud path — which was not counted at all — now is; and the two acquisition toggles are mutually exclusive and renamed to "Has acquisition data" / "No acquisition data (direct)", with the model documenting that they select on record presence rather than on a UTM tag, and the Channel cell agreeing with that definition.

Re-verified: `gofmt -l`, `make lint`, `make check-migrations`, `go test ./internal/...`, the live repository tests against a real Postgres, `pnpm typecheck` + `lint` in `web/` and `admin/`, and `pnpm build` in `site/`.

PR #365, opened 2026-09-07
feat(worker): send through every kind of SMTP server

### LINKED ISSUE #359: SMTP: a server that advertises only AUTH LOGIN cannot send, and is reported to the user as a wrong password
## Problem

`sendRaw` authenticates with `smtp.PlainAuth` unconditionally (`internal/client/smtpimap/smtp/client.go:280`). It never reads the mechanisms the server advertised, and Go's `(*smtp.Client).Auth` does not check either: it sends `AUTH PLAIN` blind.

A server that advertises `AUTH LOGIN` and not `PLAIN` therefore rejects every send. Exchange Online relays and most appliance relays are in that group. The rejection is mapped to `ErrMailInvalidCredentials`, which is CRITICAL, so the mailbox is deactivated and the drawer tells the user to check their password, which is correct.

The onboarding probe has the same limitation (`internal/email/smtp.go:65`), so the mailbox cannot be connected in the first place. Between them, a whole class of server is unusable with no way for the user to tell why.

## The fix already exists in the repo

`internal/notify/smtp.go:191-217` negotiates from `client.Extension("AUTH")`, preferring CRAM-MD5, then LOGIN, then PLAIN, with a hand-rolled `loginAuth` at `:243-281` whose comment reads "Exchange Online and many appliance relays advertise it exclusively." None of it is wired into the customer send path.

Move that implementation somewhere both can use rather than copying it, since it is the code that handles credentials.

## Also

- an AUTH failure is mapped to `ErrMailInvalidCredentials` regardless of the reply code, so a temporary 4xx auth failure or a rate-limited AUTH deactivates the mailbox as though the password were wrong
- the client never calls `client.Hello`, so `net/smtp` announces the literal `localhost` in EHLO. `internal/notify/smtp.go:144-149` overrides this deliberately, with a comment that relays treat it as a spam signal

### LINKED ISSUE #360: SMTP: no deadline after the dial, so a silent peer parks a send forever
## Problem

`sendRaw` bounds only the TCP connect, through the dialer's 10 second timeout (`internal/client/netbind/netbind.go:22`). The context is used for `DialContext` and nowhere else, and there is no `SetDeadline`, `SetReadDeadline` or `SetWriteDeadline` anywhere in `internal/client/smtpimap/smtp/`. `net/smtp` sets none of its own.

Once connected, the greeting, EHLO, STARTTLS handshake, AUTH, MAIL, each RCPT, DATA and the final `w.Close()` (where the server returns its verdict on the message) all block forever against a peer that stops answering without closing the connection. Cancelling the caller's context cannot rescue it, because nothing watches `ctx.Done()`.

This is the same failure #345 fixed on the IMAP side, on the sending side. A NAT or firewall that drops the mapping parks the send goroutine indefinitely.

## Fix

`internal/notify/smtp.go:114` and `:134-136` already do it correctly: derive a deadline from a context timeout and set it on the connection after dialing. Do the same here, and give the send path its own timeout constant so a slow relay is failed and retried rather than held.

### LINKED ISSUE #361: SMTP: a permanent rejection is retried as though the server were offline
## Problem

`sendRaw` maps almost every post-connection failure to `ErrMailServerUnreachable` (`internal/client/smtpimap/smtp/client.go:253, 260, 270, 273, 304, 320, 323, 331`), which is a Warning with resolve method Retry. The send loop then retries four times with backoff (`internal/app/worker/wmail/send.go:109`).

Two of those are permanent refusals, not outages:

- a `MAIL FROM` rejection at `:304`, which is how a blocked sender, an over-quota mailbox or a relay-denied policy arrives
- a 5xx at the end of DATA at `:331`, the server's verdict on the whole message

Only a domain-authentication refusal is rescued, and that is done by matching text in the error string (`authreject.go:28-45`) rather than by reply code. Everything else is retried three more times and then reported to the user as "the server may be offline", which is not what happened.

`ErrMailRecipientRejected` is also a Warning (`internal/errx/email.go:204-209`), so a hard 550 "no such user" on a RCPT is retried with backoff before being abandoned.

## Fix

Read the SMTP reply code, which `net/smtp` exposes as `*textproto.Error`. Treat 5xx as permanent (no retry, classified as a rejection and fed to bounce accounting) and 4xx as transient (retry). That is what the codes are for, and it removes the reliance on substring matching.

meszmate:
Closes #359, closes #360, closes #361.

#350 made receiving work on every IMAP server. This is the sending half, and the first item is the one that blocks real customers outright.

## A server that advertises only LOGIN could not send at all

`sendRaw` called `smtp.PlainAuth` unconditionally and never read the mechanisms the server advertised. Go's `Auth` does not check either: it sends `AUTH PLAIN` blind. A server offering only `AUTH LOGIN`, which is Microsoft 365 relays and most appliance relays, refused every send. The refusal was mapped to `INVALID_CREDENTIALS`, which is CRITICAL, so the mailbox was deactivated and its owner told to check a password that was correct. The onboarding probe had the same limitation, so the mailbox could not be connected in the first place.

The fix was already in the repo. `internal/notify/smtp.go` negotiates CRAM-MD5, then LOGIN, then PLAIN from the advertised list, with a hand-rolled LOGIN implementation whose comment names Exchange Online explicitly. It had never been wired into the customer send path. Rather than copy it, this moves it into the SMTP client package and has the notification mailer use it, so there is one implementation of the code that handles credentials.

An AUTH failure is also no longer read as a wrong password unless the server said so with a 5xx. A 4xx is the server declining for now, and deactivating a mailbox over it was wrong.

## Nothing bounded a send after the dial

Only the TCP connect had a timeout. The context was used for `DialContext` and nowhere else, and `net/smtp` sets no deadline of its own, so the greeting, AUTH, every RCPT and the wait for the server's verdict at the end of DATA all blocked forever against a peer that stopped answering without closing the connection. This is the same failure #345 fixed on the IMAP side. The whole conversation now carries a deadline.

## A permanent refusal was retried as an outage

Almost every post-connection failure became `SERVER_UNREACHABLE`, a warning the send loop retries four times with backoff. That included a refused `MAIL FROM`, which is how a blocked sender, an over-quota mailbox and a relay-denied policy arrive, and a 5xx at the end of DATA, the server's verdict on the message. Users were told the server might be offline when it had given a clear reason.

Refusals are now read by reply code. A 5xx is permanent: reported as `SEND_REJECTED` or `RECIPIENT_REJECTED`, carrying the server's own words, counted as a hard bounce, and not retried. A 4xx still retries, so greylisting and a busy server behave as before.

The no-retry rule is deliberately keyed on those two codes rather than on the resolve method. A general "retry only what says retry" rule would also change the Gmail and Graph paths, where `GOOGLE_FORBIDDEN` and `RATE_LIMIT_EXCEEDED` carry non-retry methods while still being retried today. That is worth looking at separately.

## Also

The client never called `Hello`, so `net/smtp` announced the literal `localhost` in EHLO. It now announces the sender's own domain, which is what the notification mailer already does with a comment noting relays read the default as a spam signal.

## Verification

- `make lint` and `go test ./internal/...` pass; `pnpm types:check` passes in `docs/`.
- The tests run against a hand-rolled in-process SMTP server, because what matters here is which mechanisms a server chooses to advertise and `net/smtp` ships no server to configure. They assert the command the client actually sent, not just that the send succeeded: LOGIN when only LOGIN is offered, PLAIN when only PLAIN is, LOGIN when both are, CRAM-MD5 when offered, and no AUTH at all against a server that advertises none.
- The permanent-versus-transient split is covered at all three points a refusal can arrive, in both directions, and the silent-peer case returns in three seconds instead of never.

## Not covered

I have not connected to a real Microsoft 365 or appliance relay. The mechanism negotiation is exercised against an in-process server that advertises what those relays are documented to advertise.

## Summary by CodeRabbit

- **New Features**
  - SMTP authentication now adapts to server-supported CRAM-MD5, LOGIN, and PLAIN mechanisms.
  - SMTP connections enforce stronger security defaults and improved timeout handling.
  - Authentication and recipient rejection errors now include clearer server details when available.

- **Bug Fixes**
  - Permanently rejected messages are no longer retried and are reported as hard bounces.
  - Temporary delivery failures continue to support automatic retries.

- **Documentation**
  - Added troubleshooting guidance for cleartext authentication refusals and rejected or temporarily failed email delivery.


meszmate:
All three were right. Fixed in the latest commit.

**Recipient rejections had no detail.** The troubleshooting row promised the server's own words for both codes, but only `SEND_REJECTED` carried them. Rather than weaken the docs I gave `RECIPIENT_REJECTED` the same treatment, because the detail is what separates an address that no longer exists from one a policy blocked, and those call for opposite responses. `TestRecipientRejectionCarriesTheServersReason` covers it.

**The cleartext refusal was classified as an outage.** Correct, and worse than it looks: `LoginAuth.Start` and `PlainAuth.Start` both refuse locally before anything reaches the server, so `permanentReply` sees no reply code, and the send was retried four times against a server that was answering fine while the operator was told it might be offline. It now maps to its own error naming the real problem, which no retry can fix.

**The race on the fake's reject map.** Also right, and it matters here because CI runs `go test -race ./...`. The canned refusal is now fixed at construction through `newRejectingServer`, so nothing writes to the map after the session goroutine starts. The package passes under `-race`.

PR #366, opened 2026-09-07
feat(worker): let a fixed mailbox problem stop showing as an error

### LINKED ISSUE #362: Mailbox errors are never cleared once the condition passes
## Problem

`email_account_errors` rows are only ever resolved by a credential reconnect: the sole caller of any resolve helper is `internal/app/email/reauth.go:207`, and it resolves credential codes only. Nothing clears anything else, and the mailbox drawer renders every unresolved row in its red "Needs attention" box (`web/src/components/app/emails/InboxDetails.tsx:659-674`), with `calculateAccountHealth` holding the mailbox at "warning" while one exists.

So a `SERVER_UNREACHABLE` written during a five minute outage in March is still shown in September, and the mailbox never returns to healthy on its own.

#350 made this more visible rather than causing it. Before, the IMAP sync swallowed transport errors by mapping them to nil, so the sync loop rarely wrote one; now a genuine outage is reported, correctly, and then never withdrawn. The two folder warnings it added (`MAX_FOLDERS_REACHED`, `FOLDER_ID_CONFLICT`) have the same problem, and both are conditions a user can fix, after which the warning should go away by itself.

## Fix

Resolve a transient error when the condition that raised it has passed:

- a sync pass that completes without a transport error resolves `SERVER_UNREACHABLE` and `CONNECTION_LOST` for that mailbox
- a folder listing that no longer overflows, or no longer has conflicting ids, resolves the matching warning
- a successful send resolves the send-side transient codes

The worker already relays `SYNC_STATE` every pass, so the consumer has a place to hang this without a new event. Note the drawer also renders a WARNING identically to a CRITICAL; severity is carried on the row (`internal/models/analytics.go:174`) and could distinguish them visually.

### LINKED ISSUE #363: The IMAP flag scan holds a full snapshot per folder, up to 74 MB per mailbox
## Problem

`imapScanFlags` (`internal/app/worker/wmail/sync_imap_flags.go`) keeps the previous scan as `map[uint32]imap.FlagState`, holding the Message-ID string and the flag slice for every message in the window, per folder, for the life of the worker process.

The window is `config.ImapFlagScanWindow` (5,000 UIDs) and a mailbox may have up to `config.MaxEmailFolders` (100) folders. Measured with a representative Message-ID and one flag:

| Approach | Per mailbox, worst case |
| --- | --- |
| Current, full snapshot | 74 MB |
| Storing only a flag hash | 14 MB |

A cold-SMTP worker's base capacity is 16 mailboxes (`worker_capacity_view`), so the worst case is roughly 1.2 GB against 230 MB. It applies only to servers without CONDSTORE, which is the Microsoft 365 and Outlook.com population the scan was written for, and those are exactly the accounts likely to have many folders.

## Fix

The previous scan only has to answer "did this message's flags change". Store a hash of the sorted flag set, `map[uint32]uint64`. The Message-ID is not needed from the previous scan at all, because the current fetch already carries it for any UID whose hash differs.

meszmate:
Closes #362, closes #363.

Two follow-ups to #350, both about it being tidy rather than merely correct.

## An error row is never withdrawn

The only caller of any resolve helper is the credential reconnect in `internal/app/email/reauth.go`, and it resolves credential codes. Nothing else clears anything, and the mailbox drawer renders every unresolved row in its red "needs attention" box while `calculateAccountHealth` holds the mailbox at "warning".

#350 made this visible rather than causing it. Before, the IMAP sync mapped transport errors to nil and rarely wrote one; now a genuine outage is reported, correctly, and then never withdrawn. A five-minute blip left a red error on the mailbox permanently.

A relayed `SYNC_STATE` is the signal that was missing. It only arrives when a pass completed, which means the worker reached the server, listed its folders and finished the tick, so it disproves exactly the connection errors and nothing else. `SERVER_UNREACHABLE`, `CONNECTION_LOST` and `RESOURCE_NOT_FOUND` are now resolved on it. Credentials, domain authentication and the fair-use deactivations are deliberately untouched: a completed sync says nothing about those, and clearing them would hide a problem the mailbox's owner still has to fix.

## The folder warnings were the wrong shape

The two warnings #350 added were events, but they describe a *state*: this mailbox has more folders than we follow, or two of its folders share an id. Both are things the user can fix, and neither could ever be withdrawn, so the warning outlived the problem.

They now ride the sync state and are rendered in the drawer's Sync card next to the deferred-message count, which is where the other "here is what is happening to your sync" lines already live. The note disappears on the next pass once the cause is gone. The two error codes and their variants are removed, since nothing raises them.

This is a better fit than the error table in a second way: the Sync card is informational, while the errors box is red and sits under "Needs attention" regardless of severity, which is a lot of alarm for "you have 105 folders and we sync 100 of them".

## The flag scan kept more than it needed

`imapScanFlags` held the Message-ID and the flag slice for every message in the window, per folder, for the life of the worker. Measured at the window and folder limits:

| Approach | Per mailbox, worst case |
| --- | --- |
| Full snapshot | 74 MB |
| Fingerprint per UID | 14 MB |

A cold-SMTP worker's base capacity is 16 mailboxes, so that is roughly 1.2 GB against 230 MB, and it applies only to servers without CONDSTORE, which is the Microsoft 365 population with the most folders. The previous scan only has to answer "did this change", and the current fetch already carries the Message-ID for any UID that did, so it now stores an order-independent digest of the flag set.

## Verification

- `make lint` and `go test ./internal/...` pass; `pnpm typecheck` and `pnpm lint` pass in `web/`, `pnpm types:check` and `pnpm lint` in `docs/`.
- New consumer tests cover both halves of the resolution rule: the connection codes are cleared, and the actionable ones are not.
- The folder test now asserts the counts are relayed as state, never as an error row, and drop back to zero once the condition clears.

## Summary by CodeRabbit

* **Bug Fixes**
  * Transient mailbox connection and server errors now clear automatically after a successful sync.
  * Sync warnings now accurately reflect folder-cap and duplicate-folder-ID issues, including affected folder counts.
  * Folder-related warnings clear when the underlying conditions are resolved.
  * Sync status warnings are shown in the mailbox drawer’s **Sync** card.

* **Documentation**
  * Updated troubleshooting and mailbox guidance to explain automatic recovery, warning locations, and folder-sync limitations.


meszmate:
All three were right.

**The stale-event race is the real one.** `NATSBus.Subscribe` creates its consumer with `MaxDeliver: 10` and no `MaxAckPending`, so out-of-order redelivery is possible, and `ResolveByCodes` resolved every matching row regardless of when it was raised. Worse than it first looks: because #350 reports an outage once rather than once per pass, a wrongly-cleared error would not have been re-raised until the next distinct outage, so the mailbox would have looked healthy while it was still down.

Resolution is now bounded to errors raised before the pass ran, through a new `ResolveByCodesBefore`. A state carrying no timestamp cannot be bounded, so it resolves nothing rather than guessing. Both cases have tests, including one asserting the bound passed is the pass's own timestamp.

**Archive** was missing from the drawer's list of folders the cap always keeps, while the docs named it. The code does keep it: `rankFolders` ranks anything `CanonicalFolder` maps away from the inbox, which includes archive. Fixed in the copy.

**The singular verb** was wrong, thank you: it read "1 folder share an internal id".

PR #374, opened 2026-09-08
Match an inbound Message-ID with and without its angle brackets

joaoppa:
## The problem

The two sides of the reply-to-campaign link disagree on the form of a Message-ID.

`advanced.cleanMessageID` strips the angle brackets off an inbound `In-Reply-To` before the lookup:

```go
func cleanMessageID(mid string) string {
    return strings.TrimSpace(strings.Trim(mid, "<>"))
}
```

`GetTaskByMessageID` then compares that bare form against the column with `WHERE message_id = $1`, while the outbound stamp always stores the bracketed RFC 5322 form. The compare could not agree, so `replied_at` was never stamped: `stop_on_reply` stayed inert and a contact who had already answered kept receiving the rest of the sequence.

## Evidence from a live self-host

A campaign test with 6 sends and 4 real replies. The replies synced with `in_reply_to` populated, and every one missed:

```
campaign | matches raw = 0 | matches normalised = 4
email    | matches raw = 0 | matches normalised = 1
warmup   | matches raw = 16 | matches normalised = 35
```

Warmup matching partly raw is the tell. `tasks.message_id` is stored bracketed 100% of the time; the variation is on the other side, because the replying client decides. On this install the synced replies split **21 bracketed to 27 bare**, so a raw compare succeeded only when the client happened to keep the brackets.

`campaign_contact_progress.replied_at` and `reply_class` were NULL for all six contacts.

## A second bug the regression test surfaced

With an empty `messageID`, the old predicate became `WHERE message_id = ''`, which is not a miss but a match: this install carries **151 tasks with an empty `message_id`**. A reply arriving with no `In-Reply-To` header at all matched one of them and linked itself to an unrelated campaign.

That is worse than the bracket bug and independent of it: not a lost attribution but a wrong one. The empty guard closes it.

## The change

- normalise the input once, then probe **both exact forms** (`WHERE message_id = $1 OR message_id = $2`, bare and bracketed)
- return early on an empty id

Probing two exact values rather than normalising the column keeps the lookup on the `message_id` index; `TRIM()` in the predicate would scan every task row on each inbound reply.

All three callers (`inbound_bounce.go:23`, `:79`, `advanced/service.go:858`) already test `task == nil`, so the early return needs no caller change.

The second commit replaces the bare `_ =` on `ProcessIncomingReply` in `HandleNewEmail` with a logged warning. It stays best-effort and non-blocking, as the comment above the call intends; only the silence changes. A failure there is currently indistinguishable from a reply that linked correctly, which is what made the bug above take a database query to find rather than a log line.

## Tests

`internal/repository/task_message_id_live_test.go`, in the existing `liveContactDB` + `WARMBLY_TEST_DB` pattern. Verified both ways: passes with the change, **fails without it**, including the `empty message id is not a wildcard` case, which is what surfaced the second bug.

gofmt, `go build ./...`, `go vet` and golangci-lint (v1.64.8) clean on top of current main.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved email reply processing by recognizing Message-IDs with surrounding whitespace or angle brackets.
  - Prevented empty Message-IDs from incorrectly matching tasks.
  - Preserved selection of the newest matching task.
  - Added error logging for incoming reply processing issues while allowing inbox ingestion to continue.

## Verified in production

Deployed on the self-host the evidence above came from, then a real reply was sent to a live campaign email:

```
replies linked today                      4 of 4
reply_class                               negative / positive, confidence 0.7, source model
follow-ups still pending for a replier    0
reply-intent warnings in the consumer     none
```

Before the change the same campaign left `replied_at` NULL on all six contacts. After it the link is made, the classifier runs, and `stop_on_reply` fires: nothing is queued for anyone who answered.

The silent consumer log also settles a question the investigation had left open. There was a suspicion that the sender-address fallback after the In-Reply-To lookup was failing independently, since it should have rescued the link even with the bracket mismatch. With the error now logged rather than discarded, it stays quiet, so there was no second failure: the bracket mismatch accounted for all of it.

### DIFF
--- a/internal/app/consumer/event_new_email.go
+++ b/internal/app/consumer/event_new_email.go
@@ -93,7 +93,14 @@ func (s *JobsService) HandleNewEmail(ctx context.Context, e *models.JobEventNewE
 	// campaign progress, gating replied_at so automated replies (auto_reply /
 	// out_of_office) never count as a human reply for stop_on_reply / branching.
 	if s.AdvancedService != nil {
-		_ = s.AdvancedService.ProcessIncomingReply(ctx, e.Message.EmailID, e.Message)
+		// Logged, not propagated: the ingest must survive it, but a silent
+		// failure here is indistinguishable from a reply that linked fine.
+		if xerr := s.AdvancedService.ProcessIncomingReply(ctx, e.Message.EmailID, e.Message); xerr != nil {
+			log.Warn().Err(xerr).
+				Str("email_account_id", e.Message.EmailID.String()).
+				Str("message_id", e.Message.MessageID).
+				Msg("Reply-intent automation failed; inbox ingest kept")
+		}
 	}
 
 	return nil
--- a/internal/repository/pg_task.go
+++ b/internal/repository/pg_task.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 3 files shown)


meszmate:
Merging this. Verified independently before doing so.

The core defect is real and is in fact wider than described. `generateMessageID` (`internal/tasks/email_sender.go:158`) always mints the bracketed form and it reaches `tasks.message_id` unmodified through `UpdateTaskMessageID`, while every caller of `GetTaskByMessageID` hands it a bare id. That is not only the reply path: the worker strips brackets at `wmail/bounce.go:47` and `:87`, and `inbound_bounce.go:18` and `:76` strip again, so inbound NDR and abuse-report attribution have been missing for the same reason. This fixes all three call sites.

I ran the regression test both ways against a live database. It passes with the change and fails on current `main` on the bare-form and surrounding-whitespace cases, so the test earns its place.

One correction for the record, since it does not change the patch but should not become repo lore: the second bug described above, an empty Message-ID acting as a wildcard, is not reachable in the shipped code. Every caller already guards it. `advanced/service.go:1042` skips an empty candidate, and `inbound_bounce.go:19` and `:77` both trim and return early. So no reply has ever reached that query with an empty id, and nothing was mis-attributed in production by that path. The wildcard really does exist one layer down (the live test confirms `WHERE message_id = ''` matches a real row), so the guard is correct defense in depth at the repository boundary and I am keeping it. It just was not a live bug.

Thanks for the fix, and for the test.

PR #375, opened 2026-09-08
Pass EMAIL_VERIFY_HELO_HOST and EMAIL_VERIFY_MAIL_FROM through to the services

joaoppa:
The in-house SMTP verifier reads `EMAIL_VERIFY_HELO_HOST` and `EMAIL_VERIFY_MAIL_FROM`, but neither is in the compose environment block, so on a Docker self-host the values sit in `.env` and never reach the container.

The prober then declines to run rather than guess a HELO name, which is the right call (a bad HELO makes the remote Postfix reject at RCPT, and that reads as "mailbox does not exist", burning good leads). But the visible result is that every contact comes back `unknown` with a message asking for a public fully-qualified HELO hostname that the operator has already set.

This adds the two passthroughs next to the other verification settings in `x-selfhost-env`. Both default to empty, so an install that does not set them behaves exactly as before.

Found on a self-hosted install where the verifier returned `unknown` for every address until the values were injected by hand.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that invalid HELO configuration skips SMTP probes, leaving email addresses with an unknown verification status.
  * Documented optional pre-send email verification, including SMTP probe settings, HELO hostname, sender configuration, fallback behavior, and backend-only scope.

* **Chores**
  * Added configuration options for external email verification through MillionVerifier, including an instance-wide API key.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -431,6 +431,24 @@ BILLING_PROVIDER=none
 # BOX_OUTLOOK_CLIENT_SECRET=
 
 
+# === Pre-send verification (optional) =========================================
+#
+# The built-in SMTP probe checks an address before a campaign sends to it. It
+# greets the recipient's server with EMAIL_VERIFY_HELO_HOST, which has to be a
+# public fully-qualified name for this instance. Unset, it falls back to the
+# host of APP_URL; when neither is a real public name the probe is skipped and
+# every address stays "unknown", which still sends.
+#
+# Backend only. Never a worker: a worker is a sending IP.
+# EMAIL_VERIFY_HELO_HOST=verify.example.com
+# EMAIL_VERIFY_MAIL_FROM=verify@example.com
+#
+# An instance-wide MillionVerifier key. Every workspace that has not connected
+# its own key is checked through this one, spending its credits, instead of the
+# built-in probe.
+# EMAIL_VERIFY_MILLIONVERIFIER_API_KEY=
+
+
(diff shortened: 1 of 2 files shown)


meszmate:
Merging. Confirmed the gap independently: on `main` there were no `EMAIL_VERIFY_*` keys in `docker-compose.yml` at all, while the backend reads them with `os.Getenv` at `cmd/backend/main.go:1795` and `:2182`, so on a Docker self-host they could never arrive. The fallback made it worse rather than softer, because `emailVerifyHeloHost()` drops back to the host of `APP_URL`, which compose defaults to `localhost`, and `isFQDN` rejects that, so `emailverify.go:285` skipped the probe and every address stayed `unknown`. `configuration.mdx` has documented all of this as settable since it was written, so the docs were describing a path that did not exist.

I pushed two things onto the branch rather than ask for a round trip:

`EMAIL_VERIFY_MILLIONVERIFIER_API_KEY` had the identical gap. It is read at `main.go:1802` and documented at `configuration.mdx:223`, and it was still not forwarded after this PR, so the verification block would have been fixed two thirds of the way. All three now go through.

A pre-send verification section in `.env.example`, since that file is where an operator goes to find a setting and it annotates which variables compose does not forward. Leaving these out of it meant the one place that answers "how do I turn this on" stayed silent.

Also trimmed the compose comment to the one-line house style and put the blank line above the group rather than below, so the SEARCH keys keep their own grouping.

PR #387, opened 2026-09-08
fix: campaign Leads tab and contact export stop partway through a large list

### LINKED ISSUE #382: One-Time Campaign: “Load More” Stops After 198 Leads
I created a one-time campaign with around 600 leads. However, when I click **“Load More,”** it only loads up to 198 leads. After that, the **“Load More”** button no longer appears, even though there are still more leads in the campaign.

Could you please check why the remaining leads are not being loaded?

@meszmate

meszmate:
Fixes #382.

## What was wrong

The contacts keyset cursor compared the sort column against a subquery that named the **outer** query's alias:

```sql
(SELECT c.created_at FROM contacts WHERE id = $n)
```

`c` is not bound inside that subquery, so Postgres resolves it as a correlated reference to the row being tested. The boundary was therefore always equal to the row's own value, and the whole clause collapsed to:

```sql
c.id >= '<cursor uuid>'
```

A UUID comparison, not a position in the sort. Each "Load more" returned the newest 50 rows of a randomly shrinking id range, so pages overlapped and the range ran dry long before the leads did. Reproduced on 617 leads: **159 distinct contacts served over 7 pages, 61 of them more than once**, then "Load more" disappeared. That matches the 198-of-600 in the report.

The same loop backs `ExportAll`, so **CSV/XLSX exports were truncated the same way**, as was `warmbly api /contacts/search --paginate`.

Sorting by `campaign_count` was worse: the cursor compared against a SELECT alias that `WHERE` cannot see, so the second page was a 500.

## The fix

`pagination.next_cursor` is still an opaque token, now `s1_` instead of `c1_`, and it carries the whole boundary: the ordering it was taken under, the boundary row's own sort value, and its id. Nothing is re-read from the row on the next page, so a lead deleted or edited mid-scroll cannot move the boundary. A token replayed under a different `sort_by`/`reverse` is a `400` rather than a silently wrong page.

Alongside it:

- the id tiebreak follows the sort direction, so one index serves both ways round
- `ORDER BY` names its NULL placement explicitly, and the keyset has NULL-aware branches behind a per-column `nullable` flag (every sortable column is NOT NULL today; the flag exists so adding a nullable one cannot silently truncate the list again)
- new index `contacts (organization_id, created_at DESC, id DESC)`, migration `000136`
- the campaign and category filters became one `EXISTS` per id instead of `IN (… GROUP BY … HAVING COUNT(DISTINCT …))`, which the planner can drive from either side
- `campaign_count` was aggregated over the **whole** `campaign_leads` table on every search, and then discarded: the response never carried it. It is now a lateral joined only when a filter or the sort asks for it

A page of a 50k-contact organization went from **120 ms to 3 ms**; the campaign Leads tab, 37 ms for the first page including the counts.

## Dashboard

- "Load more · 200 of 617", so a long list says how far through it you are
- a page that fails after rows are on screen no longer throws the loaded rows away: the failure and its retry move into the footer, and the retry re-runs the page that failed (or refetches the list when there is no next page) rather than being a button that does nothing

## Verification

`internal/repository/contact_pagination_live_test.go` pages the list the way "Load more" does and asserts every contact is served exactly once, for every sort in both directions, plus the export loop and the rejected-cursor contract. Reverting only the SQL fix turns it red with the numbers above.

End to end against a real backend, 617 leads in one campaign:

```
created_at/reverse=None      pages= 13 rows= 617 distinct= 617 total=617
first_name/reverse=True      pages= 13 rows= 617 distinct= 617 total=617
campaign_count/reverse=False pages= 13 rows= 617 distinct= 617 total=617
updated_at/reverse=False     pages= 13 rows= 617 distinct= 617 total=617
foreign cursor -> 400   old-format cursor -> 400   garbage cursor -> 400
```

CSV export of the same campaign: 617 rows, 617 distinct.

`make lint`, `make check-migrations`, `go test ./...`, and `pnpm typecheck`/`pnpm lint`/`pnpm vitest` in `web/` and `pnpm types:check`/`pnpm lint` in `docs/` all pass. Migration `000136` was applied by the real runner (`CREATE INDEX CONCURRENTLY`, version 136, not dirty).

## Docs

`api/reference/contacts.mdx` and `openapi.json`: what the cursor now carries and when it is a `400`, the real `sort_by` values, and `reverse`, which was documented backwards (it is ascending when true).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved contact list pagination so all contacts appear exactly once across sorting, filtering, reversing, and exports.
  - Invalid, malformed, or mismatched pagination cursors are now rejected with clear errors.
  - Contact tables preserve loaded results when filters hide rows and support retrying failed page loads.

- **Documentation**
  - Clarified cursor-based pagination, supported sorting, reverse-order behavior, totals, and import responses.
  - Documented that cursors are opaque and tied to the sort order used to create them.


meszmate:
@coderabbitai review

meszmate:
@coderabbitai review

PR #395, opened 2026-09-09
feat: scored worker placement and a pull-based fleet

meszmate:
## What this changes

Two connected changes to how the fleet works.

**Workers no longer have categories.** `free_tier`, `worker_type`, `risk_pool` and `egress_kind` are gone. Placement is a score over live state (capacity headroom, incumbency, region, tenant blast radius, per-provider crowding), and rotation is deliberately reluctant.

**The fleet is pull-based.** A machine joins with one command and a token, heartbeats, and asks what version it should run. The Hetzner provider, provisioning tables, worker profiles, AWS credentials and the SSH orchestrator are all removed.

## Why the categories had to go

They rested on a premise that is false for this architecture: that the worker's IP is the sending identity.

It is not. A worker never talks to a recipient MX. It authenticates to the customer's own mailbox provider, which delivers from its own outbound pool. So:

- **The worker IP is invisible to recipient spam filtering.** Google strips the submitting client's IP; Microsoft dropped `X-Originating-IP`. Co-locating a spam-prone mailbox next to a healthy one cannot contaminate the healthy one, which is why hard risk-pool segregation bought nothing.
- **The worker IP is very visible to the mailbox provider**, where it drives sign-in challenges, per-IP auth throttles (`454 4.7.0`) and per-IP rate limits (`421 4.7.28`). Exchange Online also caps SMTP AUTH at ~3 concurrent connections and ~30 msg/min per mailbox.

The practical inversion: **IP stability per mailbox beats IP diversity.** A migration is a cost, not a win. A fleet where nothing rotates is a healthy fleet.

## Onboarding

```bash
warmblyctl fleet join-token            # on the instance

curl -fsSL https://<instance>/join.sh | sh -s -- \
  --url https://<instance> --token <token> --role worker --region eu-central
```

No inbound port, no keypair, no cloud account. The join script is embedded in the binary and served by the instance itself, so a self-hosted fleet never depends on a vendor host. `PRIMARY_DB` is deliberately excluded from the config a node receives — a worker reaches relational data through the internal API and nothing else.

## Consumers are first-class

They previously had no identity at all: no id, no heartbeat, no row. A dead one was invisible until work backed up. They now enrol, heartbeat, report usage and self-update exactly like workers. A worker additionally gets a placement row; a consumer does not.

## Auto-update

The heartbeat reply carries `desired_version`; a systemd timer on the host pulls and restarts, so the process being replaced is never the process doing the replacing. Per-node pins allow canarying or holding a machine back. The backend is deliberately excluded: it is what tells everyone else their version, so a bad self-update would leave nothing to recover with.

## Bugs found along the way

- `MailboxWeight` switched on `"gmail-api"`/`"graph-api"`, which nothing ever wrote. The enum is `gmail`/`outlook`/`smtp_imap`, so every API mailbox was silently weighed 20x too heavy.
- The old rebalancer documented a 24h per-mailbox cooldown but had no column to enforce it with. There is now `worker_assigned_at`, and it is actually checked.

## Schema

- `000140` drops the four worker categories, rebuilds the capacity view.
- `000141` adds `fleet_nodes` (the registry every role shares) and reduces `workers` to a pure placement extension; `workers.id` IS the node id, enforced by a foreign key.
- `000142` drops the push-based provisioning subsystem.

All three round-trip down and back up.

## Verification

Beyond `gofmt` / `make lint` / `go vet` / `pnpm typecheck` / `pnpm lint` / `shellcheck -s sh`:

- fresh database, all 142 migrations applied, backend boots
- token issued, worker and consumer both join, both heartbeat, both told to update, `fleet list` renders the pending update
- placement places, load accounting comes out at exactly `2.05` for one SMTP + one Gmail + one SMTP mailbox (the `MailboxWeight` fix, demonstrably live)
- a healthy fleet does not churn; a worker that stops heartbeating is fully evacuated in one rotation pass, with `decision_log` entries

New live tests cover placement, rotation and the rewritten SQL: `internal/app/fleet/rotation_live_test.go` and `internal/repository/placement_queries_live_test.go`.

## Known, not caused by this branch

`TestLiveRepairedVerificationRestoresRouting` in `internal/tasks` fails on this branch. It fails identically on plain `origin/main`, so it is pre-existing and out of scope here.

## Not included

The admin Fleet section still has a `DedicatedTab` and `CapacityTab` built for the old shape. They typecheck and do not crash, but they do not yet surface the join-token flow or the version control that the CLI has.

## Deploy note

`000141` and `000142` drop columns and tables the previous release's code still reads (`workers.name`, `ip_addr`, `active`, `last_seen_at`, the SSH columns, the provisioning tables). A rolling deploy would leave old backend replicas querying columns that no longer exist.

Deploy this one stop-the-world: stop the backend and consumer, migrate, start the new build. Existing workers are carried into `fleet_nodes` by the migration and keep their identity, mailboxes and placement, so they reconnect on their next heartbeat without being re-enrolled.

Nodes running the previous worker image keep working — they just report to the old heartbeat path, which is gone, so they will show as unreachable until they are updated. Set the fleet version once after deploying and they self-update.


meszmate:
@coderabbitai full review

PR #397, opened 2026-09-09
Fix: inbox scroll position resets to top when opening emails

### LINKED ISSUE #396: Fix: Inbox Scroll Position Resets to Top When Opening Emails
In the Inbox email list, when I scroll down to find an email near the bottom and open it, the list automatically jumps back to the top.

The email list should maintain the same scroll position after opening or viewing an email, so when I go back to the inbox, I should return to the exact position where I was previously.

Image

@meszmate

meszmate:
Closes #396.

## The cause

Not the inbox: the app shell. `RouteBoundary` keyed the entire routed subtree on the pathname.

```tsx
return <ErrorBoundary key={pathname}>{children}</ErrorBoundary>;
```

The unibox puts the open thread **in the path** (`/app/unibox/all/<threadId>`), so clicking a row changed the pathname, changed that key, and React tore down and rebuilt the whole page. New DOM means `scrollTop: 0`. The same remount was also throwing away whatever the user had typed into the list search, and re-firing every loaded page of the search query on each open.

## The fix

**The shell distinguishes "different page" from "different place in this page."** New `useRouteKey` (`web/src/hooks/useRouteKey.ts`): identity defaults to the full pathname, so every other route behaves exactly as it did. A route opts params out of that identity with `handle: { stableParams: [...] }`, and the unibox declares `scope` and `threadId` as in-page state. `ErrorBoundary` now takes a `resetKey` and clears a caught error in `getDerivedStateFromProps` instead of being re-created, so navigating away from a broken page still recovers. `AppShell` keys its Suspense boundary and its content-panel scroll reset on the route identity, so a genuinely new page still mounts clean and still lands at the top.

**The list remembers where it was** — `web/src/hooks/useScrollMemory.ts`, keyed per scope + filters. This covers what keeping the page mounted cannot: leaving the inbox and coming back, and the mobile pane, which is `display: none` while a thread is open (browsers zero a hidden scroller with no scroll event). It stops on wheel, touch, key or pointer-down, gives up if the list has stopped growing short of the target, and only ever repairs a top the user did not choose.

**Reading a thread no longer refetches the list underneath you.** `useMarkSeen` patches the read/unread flip straight into the cached list and thread and refetches only the counters. It previously invalidated `["unibox"]`, so opening one email re-fetched every loaded page and re-ordered rows around whatever had arrived since.

Two related items while in there: the list auto-loads the next page as you reach the end (the button stays as the manual fallback), and `ThreadView` is keyed per thread so a half-typed reply never follows you into the next conversation.

## Worth knowing

- **Unread scope.** A thread you open now stays in the Unread list, un-bolded, until you leave the scope or the list refetches, instead of vanishing under your cursor as you read it. That is the same behaviour Gmail and Outlook have, and removing the row mid-read is the exact jump this issue is about.
- **Drive-by.** `["unibox", "count"]` was invalidated in two places but the query is `["unibox", "unseen-count"]`, so the title/favicon unread badge never moved on a snooze. Fixed both.

## Verification

- `web/`: `pnpm typecheck` clean, `pnpm lint` clean (83 warnings, all pre-existing and unchanged), `pnpm vitest run` 96/96.
- `docs/`: `pnpm types:check` and `pnpm lint` clean.
- New `web/src/app/app/unibox/uniboxScroll.test.tsx` mounts the real shell around the real unibox route. Each of its four assertions was checked to fail without the fix: reverting the boundary key breaks "keeps the list where it was" and "keeps what the user typed", and removing the scroll memory breaks both restore cases.

Not verified in a browser; that pass is still worth doing on the mobile pane in particular.

## Summary by CodeRabbit

* **New Features**
  * Unibox conversation lists now load additional conversations automatically as you scroll.
  * Your search text and list position are preserved when opening a conversation or returning to the inbox.
  * Conversations remain visible while more results load, with a manual fallback option.

* **Bug Fixes**
  * Improved message read-status updates and unread-count refresh behavior.
  * Switching conversations no longer unnecessarily resets the inbox view.

* **Documentation**
  * Updated the Unibox guide to describe automatic loading and scroll-position restoration.

### DIFF
--- a/docs/content/docs/guides/unibox.mdx
+++ b/docs/content/docs/guides/unibox.mdx
@@ -47,6 +47,8 @@ Search (press `/`) searches within your current scope, across subjects and the t
 
 Each row is a conversation, not a message, with a badge showing how many messages are inside. Opening it shows every message in order with participants and the owning mailbox. Replies and new inbound mail land in the same thread. Rows group under `Today`, `Yesterday`, `This week`, and `Earlier` headers.
 
+The list loads more conversations as you reach the end of it, and it keeps its place: opening a conversation, or leaving the inbox and coming back, returns you to the row you were on rather than the top of the list.
+
 Threading applies on both sides. A reply you send carries the provider's thread id, which nests it in your own mailbox, and an `In-Reply-To` header naming the last message in the conversation, which is what nests it for the recipient. Their mail client only ever sees the header: a thread id means nothing outside the mailbox that issued it.
 
 <Callout type="info" title="Keyboard navigation">
--- a/web/src/app/app/unibox/page.tsx
+++ b/web/src/app/app/unibox/page.tsx
@@ -19,7 +19,7 @@ import { ChevronLeftIcon, InboxIcon } from "lucide-react";
 import { ConversationList } from "@/components/app/unibox/ConversationList";
 import { ScheduledList } from "@/components/app/unibox/ScheduledList";
 import { ThreadView } from "@/components/app/unibox/ThreadView";
-import { ScopeRail, type UniboxScope } from "@/components/app/unibox/ScopeRail";
+import { ScopeRail, scopeKey, type UniboxScope } from "@/components/app/unibox/ScopeRail";
 import { ScopeSheet } from "@/components/app/unibox/ScopeSheet";
 import { UniboxHeader } from "@/components/app/unibox/UniboxHeader";
 import useFeatureAccess from "@/hooks/useFeatureAccess";
@@ -336,6 +336,7 @@ export default function UniboxPage() {
                 )}
               >
                 <ConversationList
+                  scopeKey={scopeKey(scope)}
                   scopeLabel={scopeLabel}
                   params={params}
                   setParams={setParams}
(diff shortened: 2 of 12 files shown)


meszmate:
Addressed the review on `cdbdad09`.

**Load-more failure (Major).** The runaway-retry premise doesn't hold as written — `isFetchNextPageError` is `isError && fetchDirection === 'forward'`, so a failed next page also sets `isError`, the list fell into the full-pane error branch, and the sentinel was unmounted. But that pointed at something worse: a failed `fetchNextPage` **replaced the entire loaded list** with "Couldn't load inbox". Pre-existing, except auto-loading turns it from a deliberate click into something scrolling can trigger. Fixed all three parts: the full-pane error is now reserved for when there are no rows at all, a pagination failure keeps the list and reports itself inline in the footer with a Try again, and the observer is gated on `isFetchNextPageError` so it does not retry on a loop.

**Debounce (Nitpick).** Taken, via the repo's existing `useDebouncedValue`. It was already firing a request per keystroke; the 30-minute `gcTime` in this PR is what would have made each one park a cached page, so it belongs here.

**Test isolation (Minor).** Correct, and the leak went further than `tops`: the hook's offset map is module-level and every test shared one list identity. `tops` is now rebuilt per test and each test drives its own scope (`all`, `unread`, `today`, `week`), so no assertion can be satisfied by a previous test's offset. Re-checked that each of the four still fails against the specific defect it pins.

PR #407, opened 2026-09-10
fix(imap): stop sending MODSEQ to servers without CONDSTORE, which failed every sync pass on IONOS

### LINKED ISSUE #405: IMAP sync fails every cycle with empty error (IMAP_UNKNOWN) against IONOS — connection itself is healthy
**Environment**
- Self-hosted, `install.sh` deployment (not a `git clone` dev checkout — no `Makefile` present, so `make doctor` isn't available on this instance)
- Version: `v0.4.1` (confirmed via image tags in `docker compose ps`)
- All 12 services up and healthy (backend, worker, consumer, postgres, redis, nats, realtime, tracking, forms, admin, web, updater)
- Provider: IONOS (`imap.ionos.com:993`, implicit TLS)

[manual-imap-tests-and-pcap-analysis.txt](https://github.com/user-attachments/files/32016767/manual-imap-tests-and-pcap-analysis.txt)
[email_account_errors-query-result.txt](https://github.com/user-attachments/files/32016766/email_account_errors-query-result.txt)
[worker-mail-sync-errors.txt](https://github.com/user-attachments/files/32016765/worker-mail-sync-errors.txt)
[backend-log-tail50.txt](https://github.com/user-attachments/files/32016768/backend-log-tail50.txt)
[docker-compose-ps.txt](https://github.com/user-attachments/files/32016764/docker-compose-ps.txt)

**Symptom**

SMTP works fine on this mailbox. IMAP sync fails every poll cycle (~55–65s interval) with:

```
{"level":"warn","error":"Email (<uuid>): Something went wrong: ","email_id":"<account-uuid>","message":"mail sync error"}
```

The `email_account_errors` table shows the same thing — `error_code=IMAP_UNKNOWN`, `message` field is a **genuinely empty string**, not a logging truncation. Confirmed via `strings` on `/app/worker`, which shows the literal format string:

```
"Something went wrong: %s"
```

The `%s` is being fed an empty string — so the underlying error's `.Error()` text is empty at the point Warmbly's own code wraps it.

**Ruled out, with evidence**

| Layer | Test | Result |
|---|---|---|
| Stored config | `SELECT * FROM email_accounts_smtp_imap WHERE email_account_id=...` | `imap_host=imap.ionos.com`, `port=993`, `security=tls` — correct for IONOS implicit TLS |
| Credentials | Manual `openssl s_client` + `LOGIN` from the **host** | `a1 OK LOGIN completed` |
| Full IMAP sequence | Manual `LOGIN` → `SELECT INBOX` → `FETCH 1:* (UID FLAGS)` → `LOGOUT` from the host | All `OK`, real message returned |
| Container networking | Same full sequence run from **inside the `worker` container** (after installing openssl there) | Identical clean success — TLS handshake, login, logout all fine |
| TLS/cert | Checked SAN on the presented cert | `imap.ionos.com` is listed — no hostname-mismatch issue |
| Concurrency limits | Opened two simultaneous manual sessions on the same account | Both logged in fine — not a concurrent-connection rejection |
| Live packet capture | `tcpdump` on the worker container during a real failing sync cycle | Single healthy TLS session held open 2+ minutes, real `Application Data` exchanged in both directions every ~55s (matching the poll interval), no `RST`/`FIN`/retransmits anywhere |

**Conclusion**

The network layer is doing its job every cycle — a full round trip of real data completes successfully right before the error fires. This points at something in the worker's own IMAP response handling swallowing an error *after* a successful exchange, not a connection failure. 

One possible lead: IONOS's `SELECT` response includes non-standard trailing text IMAP4rev1 servers sometimes add:

```
* OK [PERMANENTFLAGS (\Answered \Flagged \Draft \Deleted \Seen \*)] Unlimited
```

That trailing `Unlimited` is technically valid IMAP grammar but is the kind of provider quirk that can trip stricter response parsing (worker binary shows `go-imap/v2` types, e.g. `imap.idleConn`, `imap.CapSet`, `imap.NumSet`).

**What would help**

Happy to run a debug/patched build to surface the real error text if useful — the swallowed message is the main blocker to diagnosing further from the self-hosted side.

meszmate:
Closes #405.

The reporter ran v0.4.2, which made the worker print what the server actually said, and IONOS said:

```
BAD expected fetch-att instead of "MODSEQ BODY.P"
BAD expected fetch-att instead of "MODSEQ BODYST"
```

So it was ours all along. `FetchEnvelopes` asked for `ModSeq: true` unconditionally, and `MODSEQ` is a CONDSTORE fetch item (RFC 7162, and go-imap's own field comment says `// requires CONDSTORE`). IONOS advertises no CONDSTORE, so every FETCH we sent was a malformed `fetch-att` and every sync pass failed at the same point, forever, on any strict server without CONDSTORE.

The two different error strings are the same bug: go-imap emits the fetch items by ranging over a Go map, so `MODSEQ` lands in a different position each time and the server quotes whatever followed it, `BODY.P...` on one pass and `BODYST...` on the next.

## Changes

- **`MODSEQ` is gated on CONDSTORE.** The capability was already resolved post-auth into `c.condStore` and already gated `SELECT (CONDSTORE)`, `STATUS HIGHESTMODSEQ` and `SEARCH MODSEQ`. The fetch was the one place that skipped the check.
- **LIST `RETURN` options are gated on LIST-EXTENDED**, which is the same class of bug one command over. Every `RETURN` option is RFC 5258 grammar; IMAP4rev2 implies it, plain RFC 3501 does not, and SPECIAL-USE does not bring it. A server advertising the special-use folder attributes without the extended LIST would be sent a `RETURN` it cannot parse, and a BAD there costs the account its entire folder list. IONOS does advertise LIST-EXTENDED, so this is hardening rather than a second live failure. The decision moved into `listOptionsFor`, a pure function, so it can be tested across capability shapes.

## Why the existing tests missed it

This package's test server exists precisely to prove "works on a plain RFC 3501 server", and its own comment says so. But nothing had ever called `FetchEnvelopes` under it, so the one command that ignored the capability was the one command with no test.

Worse, asserting through that server is not enough on its own: it is lenient about some items a real server rejects. So the harness now records the client's half of the conversation through a proxy, and the tests assert on the command we sent rather than on whether this particular server tolerated it.

- `TestFetchEnvelopesOmitsModSeqWithoutCondStore` fetches a real message through the recorder and fails if `MODSEQ` appears on the wire. Reverting the gate fails it.
- `TestFetchEnvelopesAsksModSeqWithCondStore` covers the other direction, so gating cannot quietly cost the CONDSTORE path the cursor its incremental sync keys on.
- `TestListOptionsNeedListExtended` is a table over capability shapes, including the IMAP4rev2 implication and the two "advertised but not unlocked" cases. Removing the gate fails two of them.
- `TestFoldersSendsBareListWithoutListExtended` proves the bare LIST end to end on the wire.

## Verification

- `make lint` (golangci-lint + `check-migrations`), `gofmt -l` clean
- `go test ./internal/client/smtpimap/imap/ ./internal/app/worker/wmail/`
- Both gates mutation-tested: reverting either fails the tests that cover it

No migration, no API change, no docs change. `guides/mailboxes.mdx` already describes this behaviour correctly.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved IMAP compatibility with servers that do not advertise CONDSTORE, preventing message synchronization failures.
  - Fixed folder listing with standard RFC 3501 servers by avoiding unsupported extended LIST options.
  - Ensured optional message metadata is requested only when supported by the server.
- **Tests**
  - Added coverage for capability-dependent message fetching and folder listing behavior across different IMAP server configurations.

### DIFF
--- a/internal/client/smtpimap/imap/client.go
+++ b/internal/client/smtpimap/imap/client.go
@@ -473,8 +473,12 @@ func (c *Client) FetchEnvelopes(ctx context.Context, uids []imap.UID) ([]*Fetche
 		BodyStructure: &imap.FetchItemBodyStructure{
 			Extended: true,
 		},
-		Flags:        true,
-		ModSeq:       true,
+		Flags: true,
+		// MODSEQ is a CONDSTORE fetch item (RFC 7162). Asking a server that
+		// never advertised it for one is a malformed fetch-att, and a strict
+		// parser answers BAD and syncs nothing: issue #405, where IONOS said
+		// `BAD expected fetch-att instead of "MODSEQ BODY.P"` on every pass.
+		ModSeq:       c.condStore.Load(),
 		InternalDate: true,
 		RFC822Size:   true,
 		BodySection: []*imap.FetchItemBodySection{{
--- a/internal/client/smtpimap/imap/fetch_caps_test.go
+++ b/internal/client/smtpimap/imap/fetch_caps_test.go
@@ -1,0 +1,187 @@
+package imap
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/emersion/go-imap/v2"
+)
+
+// putMessage appends one message to INBOX and returns its UID.
+func putMessage(t *testing.T, c *Client) imap.UID {
+	t.Helper()
+	raw := "From: sender@test\r\n" +
+		"To: warmbly@test\r\n" +
+		"Subject: hello\r\n" +
+		"Message-ID: <fetch-caps@test>\r\n" +
+		"\r\n" +
+		"body\r\n"
+	cmd := c.client.Append("INBOX", int64(len(raw)), nil)
+	if _, err := cmd.Write([]byte(raw)); err != nil {
+		t.Fatalf("append write: %v", err)
+	}
+	if err := cmd.Close(); err != nil {
+		t.Fatalf("append close: %v", err)
+	}
+	data, err := cmd.Wait()
+	if err != nil {
+		t.Fatalf("append wait: %v", err)
+	}
+	return data.UID
+}
+
+// Issue #405. MODSEQ is a CONDSTORE fetch item (RFC 7162), and this package
+// asked for it on every FETCH regardless of what the server advertised. A
+// lenient server takes it anyway, which is why every test here passed; IONOS
+// answered `BAD expected fetch-att instead of "MODSEQ BODY.P"` and the
+// mailbox synced nothing, every pass, for as long as it was connected.
+//
+// go-imap emits the fetch items from a map, so MODSEQ lands in a different
+// position each time and the failure is not even reproducible in the same
+// shape twice. Assert on the command, not on this server's tolerance.
+func TestFetchEnvelopesOmitsModSeqWithoutCondStore(t *testing.T) {
+	c, wire := recordingServer(t, nil)
+	if err := c.Connect(); err != nil {
+		t.Fatalf("Connect: %v", err)
+	}
+	if c.HasCondStore() {
+		t.Fatal("the test server must not advertise CONDSTORE")
+	}
+	uid := putMessage(t, c)
+	if _, err := c.SelectForSync("INBOX"); err != nil {
+		t.Fatalf("SelectForSync: %v", err)
+	}
+
+	got, xerr := c.FetchEnvelopes(context.Background(), []imap.UID{uid})
+	if xerr != nil {
+		t.Fatalf("FetchEnvelopes: %s", xerr.Message)
+	}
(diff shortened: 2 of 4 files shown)


akminmkannan:
Thanks — #407 confirms the root cause exactly.

I can see that the fix is now merged into `main` and that `FetchEnvelopes` was indeed requesting `ModSeq: true` unconditionally on servers without `CONDSTORE`.

My current self-hosted instance is still on `v0.4.2`, so it does not yet contain #407.


```text

akminmkannan:
able to pull v0.4.3 and this issue seems to be fixed and working fine. Thank you very much..

PR #411, opened 2026-09-10
fix: stop forcing prompt=consent on Microsoft 365 mailbox sign-in (#409)

### LINKED ISSUE #409: Outlook/M365 OAuth flow sends prompt=consent unconditionally — blocks non-admin users with AADSTS90095 even after tenant-wide admin consent is granted
### Title

Outlook/M365 OAuth flow sends `prompt=consent` unconditionally — blocks non-admin users with AADSTS90095 even after tenant-wide admin consent is granted

### Body

**Environment**
- Self-hosted, v0.4.3
- Microsoft Entra ID (Azure AD) tenant, multiple verified domains under one tenant
- Warmbly's registered Entra app: `BOX_OUTLOOK_CLIENT_ID` in `.env`

**Symptom**

Non-admin users connecting an Outlook/M365 mailbox in Warmbly hit Microsoft's "Approval required" screen and fail with:

```
AADSTS90095: Admin consent is required for the permissions requested by this application.
An admin consent request may be sent to the admin.
```

This happens **even when tenant-wide admin consent has already been granted** for the app in Entra ID (all requested Microsoft Graph scopes showing `AllPrincipals` / admin-consented). One admin user account connects successfully every time; every non-admin user account fails every time, regardless of how the tenant-wide consent grant is configured or rebuilt.

**Root cause**

The authorize URL Warmbly generates includes `prompt=consent`:

```
https://login.microsoftonline.com/common/oauth2/v2.0/authorize?access_type=offline&client_id=<client_id>&prompt=consent&redirect_uri=<redirect_uri>&response_type=code&scope=openid+email+profile+offline_access+https%3A%2F%2Fgraph.microsoft.com%2FUser.Read+https%3A%2F%2Fgraph.microsoft.com%2FMail.Send+https%3A%2F%2Fgraph.microsoft.com%2FMail.ReadWrite&state=<state>
```

`prompt=consent` forces Microsoft to present an interactive consent prompt on **every** sign-in, regardless of whether a valid admin consent grant already exists for the tenant. For scopes that require admin consent (`Mail.Send`, `Mail.ReadWrite`, etc.), only a user holding admin rights can personally satisfy that forced prompt on the spot. A non-admin user is shown the "Approval required" wall every single time — the pre-existing tenant-wide grant is never checked because `prompt=consent` skips Microsoft's normal silent-token-issuance path entirely.

This matches Microsoft's own documented behavior, confirmed via a Microsoft Learn Q&A describing the identical symptom for a different third-party SaaS app:
https://learn.microsoft.com/en-us/answers/questions/5927650/error-90095-for-users-on-non-primary-verified-doma

Relevant excerpt from a Microsoft moderator's answer on that thread: they identify `prompt=consent` in the authorize request as the primary suspect for exactly this symptom — admin consent already granted tenant-wide, yet non-admin users still blocked with AADSTS90095 — and recommend removing it or comparing it against the granted scope set.

**Evidence this is not an Entra/tenant misconfiguration**

Extensive tenant-side troubleshooting was performed before finding this, all of which came up clean and pointed away from the tenant:
- Tenant-wide `AllPrincipals` admin consent grant confirmed correct and complete via direct Microsoft Graph API inspection (`Get-MgOauth2PermissionGrant`), matching all scopes in the authorize request
- Grant deleted and rebuilt via the Entra portal, still failed for non-admin users
- Individual `Principal`-scoped grant manually created for a specific failing user, cloned in shape from a known-working admin user's grant, verified attached to the correct `PrincipalId` via `Get-MgUser` — still failed
- Fresh interactive consent performed through the real OAuth flow (not portal-generated), with "Consent on behalf of your organization" explicitly checked, producing a verified-correct `AllPrincipals` grant — admin user succeeded immediately with this grant, non-admin user failed immediately after, with the same grant in place

- No Conditional Access policies applied (`conditionalAccessStatus: notApplied`, 0 failure-causing policies)
- Assignment requirement set to No; also tested with explicit user assignment
- Full Exchange Online licensing confirmed (`EXCHANGE_S_STANDARD`, `ProvisioningStatus: Success`)
- Only one Entra service principal exists for the app (no duplicate registration)
- Reproduced across multiple browsers, incognito sessions, and a fresh browser profile that had never authenticated to the tenant before
- Reproduced identically for two separate non-admin users, on two separate verified domains within the same tenant

All of the above matches the pattern in the Microsoft Learn thread precisely, where Microsoft's own support staff concluded the fault was in the third-party app's OAuth request (specifically `prompt=consent`), not in Entra tenant configuration.

**Suggested fix**

Remove `prompt=consent` from the Outlook/M365 authorize request, or replace it with `prompt=select_account` (to still allow account switching without forcing a consent re-check). Once removed, Microsoft should silently issue tokens for users covered by an existing tenant-wide admin consent grant, instead of forcing every non-admin user through a prompt only an admin can complete.

If `prompt=consent` was added intentionally (e.g. to force re-consent after a scope change), consider making it conditional — only include it on first-time connection or after a detected scope change, not on every authorize call.

**Impact**

This affects any self-hosted Warmbly instance connecting Outlook/M365 mailboxes where the mailbox owner is not a tenant admin — which is the normal case for most organizations. Currently the only workaround is having a tenant admin personally complete the OAuth flow for every mailbox, which isn't viable for teams onboarding multiple mailboxes.

Image

Image

Image

Image

Image

meszmate:
Fixes #409.

## What was wrong

Every mailbox authorize URL was built with `oauth2.ApprovalForce`. In `golang.org/x/oauth2` v0.36.0 that constant is `SetAuthURLParam("prompt", "consent")`, so the Outlook / Microsoft 365 request carried `prompt=consent` on every connect and every reconnect.

Microsoft treats that as an instruction to re-run the consent eligibility check for the signing user rather than to honour the consent grants already on the tenant. A non-admin asking for `Mail.Send` / `Mail.ReadWrite` fails that live check and is stopped at "Approval required" with `AADSTS90095`, no matter how the tenant-wide `AllPrincipals` grant is built. Microsoft's own guidance on this is explicit: *"Once the application obtains consent, make sure the `prompt` parameter isn't specified. Otherwise, users might always receive a consent error."*

The parameter was never doing anything for Microsoft in the first place. Entra ID issues a refresh token off the `offline_access` scope, which is already in the request; `prompt=consent` only bought the failure. `access_type=offline` is a Google parameter and equally pointless there.

Google is the opposite case, so it is untouched: it issues a refresh token only under `access_type=offline`, and re-issues one on a repeat authorization only when the consent screen is forced. Dropping either would connect a Gmail mailbox that stops sending an hour later.

## The change

`authCodeOptions(provider, loginHint)` in the new `internal/app/email/oauth_params.go` is now the one place the authorization parameters are decided, used by all three flows that build one (`OAuthStart`, `OAuthReauth`, and the brokered `OAuthAuthorizeURL` behind Warmbly Cloud mailboxes):

| | Google | Outlook / Microsoft 365 |
|---|---|---|
| before | `access_type=offline`, `prompt=consent` | `access_type=offline`, `prompt=consent` |
| after | `access_type=offline`, `prompt=consent` | `prompt=select_account` |

`prompt=select_account` rather than no prompt at all: with no `prompt`, Microsoft signs in the sole account already in the browser session, which on a shared or already-signed-in browser silently connects the wrong mailbox. `select_account` keeps the picker and does not trigger the eligibility re-check. The reauth flow still passes `login_hint` so the mailbox being renewed is offered first, and the finish leg still refuses tokens issued for any other address.

## Docs

- `development/deployment-guide.mdx`: the Entra registration step listed `IMAP.AccessAsUser.All` and `SMTP.Send` among the permissions to add. Neither is requested any more (Graph is the transport for send and sync), and both are admin-consent-only, so following that step turned a sign-in every user could complete into one only a tenant admin could. Corrected to the four delegated permissions actually asked for, with a callout on how a tenant that has disabled user consent grants it once.
- `development/troubleshooting.mdx`: a row for the "Approval required" / `AADSTS90095` symptom.

## Verification

- `go test ./internal/app/email/` — `TestAuthCodeOptions_OutlookNeverForcesConsent` asserts the Outlook query has `prompt=select_account` and no `access_type`; `TestAuthCodeOptions_GoogleKeepsOfflineConsent` asserts Google is unchanged; `TestAuthCodeOptions_LoginHintOnlyOnReconnect` asserts `login_hint` appears only on a reconnect. Mutation-checked: reverting the Outlook branch fails the first test on both assertions.
- `make lint` (golangci-lint + gofmt + check-migrations) clean.
- `pnpm types:check` and `pnpm lint` clean in `docs/`.

## Sources

- [Microsoft Q&A: prompt=consent blocks non-admin users even when tenant-wide admin consent has been granted](https://learn.microsoft.com/en-us/answers/questions/5847660/bug-report-prompt-consent-blocks-non-admin-users-e)
- [Microsoft Q&A: Error 90095 for users on non-primary verified domains despite tenant-wide admin consent granted](https://learn.microsoft.com/en-us/answers/questions/5927650/error-90095-for-users-on-non-primary-verified-doma)
- [OAuth 2.0 authorization code flow](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow) for the `prompt` values and for "`refresh_token` ... Only provided if `offline_access` scope was requested"

## Summary by CodeRabbit

- **Improvements**
  - Outlook/Microsoft 365 mailbox connections now request only the permissions needed for reading and sending mail.
  - Outlook sign-in no longer forces repeated consent prompts and keeps the account picker available.
  - Reconnecting a mailbox can automatically select the previously used account.
  - Google mailbox authorization continues to support offline access and consent for reliable reconnection.

- **Documentation**
  - Added guidance for resolving Outlook “Approval required” errors, including tenant-wide administrator consent requirements and upgrade recommendations.


akminmkannan:
Yes.  We are able to connect to microsoft account as users now. thank you.

PR #420, opened 2026-09-10
Worker capacity is a placement target, not a hard gate

meszmate:
## What

`PlacementCandidate.Eligible` refused a worker when `Effective - Load < Weight`. That was the only hard capacity constraint, and it was doing more harm than good.

`base_capacity` is a hardcoded `16` in `worker_capacity_view` (`000141_fleet_nodes.up.sql:109`), identical for a 32-core box and a small VM. Nothing measures what a worker can actually carry, so refusing on it was refusing on a guess.

Worse, it refused at the moment placement mattered most. When every candidate sat at the target, `SelectPlacement` returned nil and `assignWorker` fell through to `selectFallback` (`assignment.go:277`), which takes the first healthy worker off the list and reads none of region match, tenant blast radius or provider crowding. A cliff that discards the scoring is worse than no cliff.

## Why the premise was wrong to begin with

The premise behind a per-worker mailbox cap is that the worker's IP is the sending identity. It is not, and migration `000140` already removed the four worker categories that rested on that. A worker never talks to the recipient's MX: it authenticates to the customer's own mailbox provider, which delivers from its own outbound pool. Google strips the submitting client's IP and Microsoft dropped `X-Originating-IP`, so recipient-side filtering never sees us.

What the worker IP *is* visible to is the mailbox provider, where it drives sign-in risk challenges and per-IP auth throttles (`454 4.7.0`, `421 4.7.28`). That is a crowding-and-pushback problem, which the score already models, not a headcount problem that wants a fence.

## Changes

**`Eligible` is health-only.** Heartbeating, `healthy` or `watch`. Capacity leaves the hard path entirely.

**Headroom is projected.** `(Load + Weight) / Effective` rather than current utilization. `req.Weight` was previously read *only* by `Eligible`, so dropping that check without this would have left the score blind to whether it was placing a `1.0` `smtp_imap` mailbox or a `0.05` Gmail one. The same nearly-full worker is now correctly under target for one and over for the other.

**Over-target is a soft wall.** A flat `weightOverTarget = 3.0` plus a `weightOverload = 4.0` ramp on the overage. The flat cost is set deliberately above the largest bonus any candidate can earn (incumbency `2.0` + region `0.6`), so a worker with room wins whenever one exists, including against a sticky incumbent -- the contract `TestIncumbentLosesWhenItHasNoHeadroom` asserted is preserved. The ramp keeps ranking overloaded workers against each other once nothing has room, which is precisely the case the hard check handled by giving up.

**Auth pressure is now scored.** `WorkerCapacityRow` has carried `AuthErrors1h` since it was written and `ComputeCapacity` never read it. Auth errors are the one signal in `worker_health_samples` genuinely about the worker's own address rather than about the mailbox or its contact list. `computeAuthPressure` scales the hourly rate into `[0, 1]` (full scale at 5%, denominator floored at 1 so sync auth failures on a worker that sent nothing saturate rather than divide by zero), weighted `1.2`.

That weight is below incumbency on purpose: evacuating a throttled worker is rotation's call, behind `RotationElevated`'s 6h residency floor, and paying a per-mailbox sign-in challenge to escape a transient throttle is how a fleet churns itself.

## Consequences

- `SelectPlacement` now returns nil only when nothing is healthy. A full-but-healthy fleet places the mailbox instead of dropping to the fallback path.
- `selectFallback` becomes what its comment already claims it is: the no-capacity-view path.
- `PlacementResult.IncumbentEligible` narrows to "healthy enough to host". It is set but unread outside tests, so no behaviour changes; comment updated.
- `Capacity.Utilization` is untouched, so the rotation hot-worker query, `fleet/scale.go` and the admin fleet panel all keep meaning what they meant.

## Not in this PR

`health_multiplier` in `worker_capacity_view` is built from bounce and complaint rates:

```sql
1.0 - LEAST(0.5, bounces_hard_1h/sends * 5)
    - LEAST(0.5, complaints_1h/sends * 100)
```

Those are properties of the mailbox and its contact list, not of the worker's address -- the same premise `000140` was meant to have killed. Today a customer with a bad list shrinks their worker's capacity, pushing it past `RotationHotUtilization`, rotating mailboxes to a new IP for a sign-in challenge that fixes nothing, because the bounces travel with the mailbox.

Left alone here: the `base * health * age` formula is duplicated across `capacity.go`, `pg_admin_fleet.go`, `fleet/scale.go` and twice in SQL in `pg_worker_placement.go`, and correcting it means a migration recreating the materialized view. It is also strictly less harmful after this PR, since a shrunk target now costs score instead of causing a refusal.

## Tests

- `TestEligibleRequiresHealthOnly` replaces `TestEligibleRequiresHealthAndHeadroom`; asserts an over-target worker is still a legal home.
- `TestSelectPlacementReturnsNilOnlyWhenNothingIsHealthy` -- a full fleet is not an empty one.
- `TestOverTargetWorkerIsLastResort` -- over-target loses even holding incumbency and a region match.
- `TestAmongOverloadedWorkersTheLeastOverloadedWins`.
- `TestMailboxWeightCountsAgainstTheTarget` -- same worker, opposite verdict for Gmail vs `smtp_imap`.
- `TestAuthPressureSteersAwayFromAThrottledAddress`.
- `TestComputeAuthPressure` table, including the zero-attempts case.

Docs: `architecture.mdx` placement table and constraints paragraph, and the Worker Topology section of `AGENTS.md`.

Verified: `make lint` clean, `gofmt -l` empty, `go test ./internal/app/worker/...` green, `pnpm types:check` clean in `docs/`.


meszmate:
Self-review found six issues. All fixed in fa2b5330.

**Auth pressure removed entirely.** The term justified itself as "the one signal genuinely about the worker's address", and it was measuring the wrong column. `recordSendOutcome` (`health_record.go:35`) routes `AuthenticationFailed` / `AuthorizationFailed` / `InvalidCredentials` / `GoogleAuth` to `RecordAuthError`, which are per-mailbox credential problems: a revoked token, a changed password. The `454`/`421` per-IP throttles it claimed to measure go to `RecordRateLimitError` and land in `rate_limit_errors`, which `worker_capacity_view` does not aggregate at all. One mailbox with a dead OAuth token on a quiet worker would have saturated the term and docked a healthy worker `1.2` on every placement.

Exposing `rate_limit_errors` needs a migration recreating the materialized view, which is the same migration the `health_multiplier` fix needs. Both are now follow-up rather than bolted on here.

**Age damping no longer divides the over-target ratio.** `age_multiplier` ramps over 72h, so a node enrolled an hour ago has `Effective = floor(16 * 1/72) = 0`, floored to `1`. Scoring `(Load + Weight) / Effective` against that made it read `-7.000` after a single mailbox, against `-4.250` for a mature worker sitting at 125% of target. New nodes would have stopped attracting work for roughly a day, and the old `selectFallback` escape (ordered by `account_count`, so it picked exactly that new node) is unreachable now. That defeats "the fleet is full, join a worker", which is the remedy this PR is built on.

`ComputeCapacity` now also produces `Capacity.Target` = `Base * HealthMul` with no age term, and placement divides by that. Health still shrinks the target, because that is about the worker. Youth becomes an explicit `weightNewWorker = 0.5` score term, small enough that it can never outweigh the over-target cost. `TestNewNodeRelievesAFullFleet` pins it.

**Isolated egress is bounded again.** The mandated branch (`assignment.go:225`) bypasses `SelectPlacement`, so the over-target cost never applied to it, and `WorthMoving` short-circuits on `Mandated`. `Eligible`'s capacity check was the only thing holding it; removing that let an entitled org put unlimited mailboxes on one reserved box. It now checks `OverTarget` explicitly and falls through to scoring when the reserved worker is full, where `weightIsolation` still prefers it. That is the documented intent anyway: a strong preference, not a pin.

**Rotation pile-on capped.** These opportunistic capacity moves are newly reachable: previously the headroom term clamped at 0, so an over-target incumbent still scored `+2.0` on stickiness and never cleared `RotationMinScoreGain`. Now the gain is ~2.69 and they fire. But `worker_capacity_view` is materialized and refreshed on its own minute cadence (`cmd/backend/main.go:1051`), so all 50 mailboxes in a tick score against one frozen `load_score` and the emptiest worker looks equally empty to every one of them. Added `MaxMovesPerDestination` (default 5).

**Two doc corrections.** The "flat cost exceeds any bonus" claim was overstated: it exceeds the bonuses (incumbency `2.0` + region `0.6`), so stickiness can never rescue an over-target worker, but it does not dominate the penalty terms, so a worker with room that is crowded with foreign tenants can still lose. Corrected in `placement.go`, `AGENTS.md` and `architecture.mdx`; `TestOverTargetLosesToStickinessButNotToEveryPenalty` asserts both halves. And `pg_worker_placement.go:98` still described app code as filtering on capacity headroom.

`make lint` clean, `gofmt -l` empty, `go test ./internal/app/worker/... ./internal/app/fleet/...` green, `pnpm types:check` clean in `docs/`.

meszmate:
@coderabbitai review

The first pass was rate-limited and only covered 7fa4fdfb. fa2b5330 is the substantive commit: it drops the auth-pressure term, moves the over-target ratio onto an age-free `Capacity.Target`, bounds the isolated-egress override with an explicit `OverTarget` check, and caps rotation moves per destination. Please review the head.

PR #441, opened 2026-09-11
feat: per-step open, click, reply and bounce rates in Step performance

### LINKED ISSUE #434: Add Individual Open Rate in percentage for Each Campaign Step
In the Step Performance section, each campaign step should display its own open rate, rather than only showing the total number of opens.

Image

Image

This would make it much easier to understand which step/email is performing best.

@meszmate

meszmate:
Closes #434.

Step performance showed counts only, so the four steps in the report (337 sent, 285, 11, 1) could not be compared: 101 opens on a step that reached 285 contacts and 91 on a step that reached 337 look like the same result until you divide them yourself.

Each count now carries its own rate underneath it, and the rate is against **that step's** sends, not the campaign's, which is what makes a fourth touch that reached 11 contacts comparable to a first touch that reached 337.

### What changed

**Backend.** `models.SequenceStats` gains `open_rate`, `click_rate`, `reply_rate` and `bounce_rate` (percentages of the step's own `emails_sent`, via the existing divide-by-zero-safe `models.Rate`), plus `machine_opens` and `machine_clicks` so a step's open rate can be read against how much of it was an Apple MPP prefetch or a security gateway. The same shape `ABVariantStats` already had. The automated-click expression is now one `machineClicksCount` constant shared by the campaign summary and the per-step query, so the two can never drift apart on what an automated click is. `GetSequenceStats` also checks `rows.Err()`, which it was not doing: a mid-iteration failure returned partial step stats as if they were the whole campaign.

**Dashboard.** Step performance keeps its six columns and gains a second line per cell: the count, then the share of that step's sends. A step that has sent nothing shows a dash, never `0.0%`, because 0% reads as a result. Where some of the events were automated the rate is dotted-underlined and the hover gives the split, matching the `N auto` note the campaign totals already carry.

**API surface.** The new fields flow through `GET /analytics/campaigns/:id` (additive, inside the version), the `warmbly` and `warmblyctl` pass-through renderers, and `get_campaign_stats`, which previously dropped the step breakdown entirely so the assistant could not answer "which touch is working".

**Docs.** `api/reference/analytics.mdx` (example plus the denominator rule), `docs/public/openapi.json` (`CampaignStepStats`), and `guides/analytics.mdx`.

### Market check

Instantly ships a dedicated Step Analytics section per campaign; Smartlead exposes per-`sequence_number` stats and per-step rates through its API and MCP tools. Both treat per-step rates as table stakes. One thing deliberately not copied: Instantly's campaign open rate divides by "sequences started" (leads who received step 1), which makes a follow-up's rate a fraction of a number it never sent to. Per-step here divides by the step's own sends.

### Verification

- `internal/repository/campaign_step_rates_live_test.go`: a campaign with an intro step (5 sent, 3 opens of which 1 is a machine open, 1 human click, 1 contact whose only click was a gateway, 1 reply, 1 bounce) and a follow-up queued but never sent. Asserts the counts, the rates (60/20/20/20), that the unsent step's rates stay 0 rather than dividing by zero, and that the steps add up to the summary. Every assertion was mutation-tested: dropping the rate computation, using the wrong denominator, and flattening either machine count each fail it.
- `make fmt`, `make lint` (golangci-lint + check-migrations), `pnpm typecheck` / `pnpm lint` / `pnpm vitest run` in `web/`, `pnpm types:check` / `pnpm lint` in `docs/`. No migration.

## Summary by CodeRabbit

- **New Features**
  - Campaign analytics now include per-step send, open, click, reply, and bounce rates.
  - Step metrics are calculated against each step’s own sends, with automated opens and clicks identified separately.
  - Campaign step-performance tables display counts, percentages, and automated-activity indicators.
  - Analytics tools and API responses now expose detailed step-level metrics, including automated activity counts.

- **Documentation**
  - Updated analytics guides and API reference documentation to explain step-level rates, automated activity, and no-send behavior.


meszmate:
Follow-up in 35d9d3d: the automated-click count is now one rolled-up `bool_and(machine)` per (step, contact), joined into both the summary and the per-step query, rather than a correlated `EXISTS`/`NOT EXISTS` pair evaluated per progress row.

The per-step stats needed the same count the summary had, and duplicating the correlated form would have paid for it twice on every campaign overview load. Measured on a seeded 50k-lead campaign with 20k logged clicks (4 steps, 12.5k contacts):

| | EXISTS pair | rolled-up join |
|---|---|---|
| automated-click count alone | 78ms (75-82 over 5 runs) | 33ms (32-34) |
| whole per-step query | — | 55ms, against 41ms for the query before this PR |

Verified identical on the same dataset: both formulations return `{tc:12500, sent:50000, op:20067, mop:2008, mcl:5929}`. The live test also asserts the steps still add up to the summary, so the two cannot drift apart on what an automated click is.

meszmate:
All three CodeRabbit findings verified and fixed in 470654f:

1. **`machine_clicks` counted "steps", not contacts** — correct, and the wording predates this PR. The count is over `campaign_contact_progress` rows, so it is the contacts whose only clicks on a step were automated. Fixed in the API reference and in the two Go doc comments that said the same thing (`CampaignSummary.MachineClicks`, `DashboardOverallStats.MachineClicks`), so the prose and the OpenAPI schema now agree.
2. **Zero-send rule only on `open_rate`** — added to `click_rate`, `reply_rate` and `bounce_rate`.
3. **`get_campaign_stats` dropped the machine counts** — added `machine_opens`/`machine_clicks` per step, and to the campaign totals as well: without them a 35% open rate reads as all-human at both levels, which is exactly the misreading those fields exist to prevent.

meszmate:
@coderabbitai review

PR #444, opened 2026-09-12
fix: make Edit with AI apply the instruction instead of rewriting the body into a fresh cold email

### LINKED ISSUE #432: AI generated email content is not edited in body
Image

When I change the email content, it does not change the email content that generated

@meszmate

meszmate:
Fixes #432.

## What was wrong

`/generation/edit` (the **Edit with AI** pill) ran on `generation.BuildVoiceRules` — the *cold-outreach writer* system prompt. It opens with "You are Warmbly's cold-outreach email writer", caps output at 80 words / 4-6 sentences, pins a five-part skeleton (observation → pain → claim → one ask → optional P.S.) and ends with "SELF-CHECK … If any answer is no, rewrite."

That system prompt outranked whatever the user typed. "Fix grammar" or "shorten this" came back as a freshly generated cold email in that shape, and every rerun converged on the same copy — which is exactly "it does not change the email content that generated". The body in the reporter's screenshot is that skeleton line for line.

Four more defects in the same flow, each reproduced against the real editor before being fixed:

| | Effect |
|---|---|
| passage read with `doc.textBetween` | atoms return `""`, so every `{{.FirstName}}`, AI block, condition and form link was deleted from the passage sent, and the rewrite came back with them gone |
| link marks never serialized | a hand-placed `{{.UnsubscribeLink}}` anchor simply stopped existing after a rewrite |
| `"\n"` out, split on `\n{2,}` back in | a multi-paragraph selection collapsed into one |
| replacement via `insertContentAt` | it closes the slice it parses, so rewriting a phrase inside a sentence produced `<p>One </p><p>REWRITE</p><p> four.</p>` |
| `writingMaxTokens = 1024` | a long passage rewrite truncated mid-sentence |

Plus the layout in the screenshot: the review row wrapped `· 1 credit · 1.7k tok` into four lines, and the card was clamped to the viewport rather than to the editor, so it hung ~60px outside the step drawer over the flow canvas.

## What changed

**Backend.** `internal/pkg/generation/voice.go` adds `editRules` + `BuildEditRules`: an *editor* frame (apply the instruction, change nothing else, return only the passage, keep length/paragraphs/language, preserve every token and link destination character for character, return it unchanged when the instruction already holds), keeping the humanizer's bans and the org grounding. `generation_edit.go` runs it through `AIProvider.Complete` with a 3072-token cap, the way reply drafts and AI variables already do, and counts its request limits in runes so the same body is not editable in English and refused in Cyrillic.

**Frontend.** New `richTextPassage.ts` owns the round trip: `passageText` walks the range with TipTap's text serializers and folds link marks in as markdown, `passageHTML` re-chips what comes back (an AI block only when its config came with the passage, a link only when the destination is one), `replacePassage` swaps the range with paste semantics and returns `null` when the document is genuinely unchanged. Wired into both `RichTextAIEdit` and `RichTextAICaret` (⌘J had the same paragraph-splitting bug). New `floatingBounds.ts` clamps the card to the surface it is editing. The review row is one non-wrapping line with the cost underneath, and says **No change** instead of **Rewritten** when the model returned the passage as it was.

**Docs.** `guides/unibox.mdx` states what an edit does and does not touch; `guides/sequences.mdx` gains a "Writing and editing with AI" section (the campaign editor's AI surfaces were not documented at all). Both say plainly that bold/italic inside a rewritten selection is not preserved.

## Verification

- `make lint` clean, `gofmt -l` empty, `go test ./internal/api/handler ./internal/pkg/generation` ok
- `web`: `tsc -b` clean, `eslint` 0 errors, `vitest` 182 passed — 28 of them new
- the frontend tests drive the real `RichTextEditor` in its controlled host, so they fail on the actual defects rather than on a stand-in
- every assertion was mutation-tested: putting `textBetween`, `insertContentAt`, `!tr.docChanged`, plain-text detokenising, `BuildVoiceRules` or `MaxTokens: 1024` back each fails the specific test that guards it

## Not in scope

`/generation/write` still uses the writer prompt as its system message. The ⌘J caret hits the same framing conflict there, but that endpoint also backs the compose window's subject generation, so reframing it is an API-shape decision rather than a bug fix.

## Summary by CodeRabbit

- **New Features**
  - Added AI writing at the cursor and **Edit with AI** for selected text.
  - AI edits preserve merge fields, conditions, spintax, AI blocks, form links, and link destinations.
  - Rewrites now apply only the requested instruction; bold, italic, and heading formatting in rewritten selections is not preserved.
  - Added **No change**, **Undo**, and **Again** feedback and actions.
  - Improved paragraph, link, and content insertion behavior during AI edits.
  - AI cards now remain within the campaign panel and viewport.

- **Documentation**
  - Updated Sequences and Unibox guides with AI writing and editing guidance.


meszmate:
@coderabbitai review

PR #452, opened 2026-09-12
fix: realtime cannot complete either of its TLS connections

meszmate:
`warmblyctl status` against production reports the realtime service unreachable, and its log explains why: it never finishes connecting to anything. Two separate TLS problems, each of which alone keeps the service from working.

## Redis: a wildcard certificate is rejected

Ten `CLIENT ALERT: Fatal - Handshake Failure - {:bad_cert, :hostname_check_failed}` per attempt, one per pool connection.

Erlang's default hostname check does not match a wildcard certificate. Every managed Redis presents one: Upstash serves `CN=*.upstash.io`, and ElastiCache and Redis Cloud do the same. Redix's own docs call this out:

> Some Redis servers, notably Amazon ElastiCache, use wildcard certificates that require additional socket options for successful verification

`Realtime.Redis` passed no socket options, so the whole pool failed to connect. Because the module fails open by design, this surfaces as "Redis unavailable" rather than as an error anyone would chase: rate limiting silently returns defaults and presence state has nowhere to live.

The fix is the documented option, applied only for `rediss://`. Redix drops just the keys you supply from its own defaults, so `verify: :verify_peer` and the system CA store still apply; this does not weaken verification, it makes it able to succeed.

## Postgres: the RDS root is not publicly trusted

`Postgrex.Protocol failed to connect: tcp connect (...rds.amazonaws.com:5432): timeout`.

The timeout is a network problem outside this repo (the service had no static egress IP, so the database security group dropped it). But fixing the network only exposes the next failure: Postgrex defaults to `[cacerts: :public_key.cacerts_get(), verify: :verify_peer, ...]`, and Amazon RDS presents a chain rooted in an RDS CA that is in no public trust store. The same problem the Go services already solved by shipping AWS's bundle and pointing `sslrootcert` at it.

Realtime is Elixir and reads no DSN parameters, so it gets the equivalent knob:

- the image now ships `/etc/ssl/rds/global-bundle.pem`, as `backend.Dockerfile` and `consumer.Dockerfile` already do
- `DATABASE_SSL_CA_FILE` points Postgrex at a bundle, with `verify_peer` and the https match fun

It is opt-in rather than the default for exactly the reason the backend's `sslrootcert` is: defaulting every install to an RDS-only store would break a Postgres fronted by a public CA. `DATABASE_SSL=true` with no CA file keeps today's behaviour.

## Tests

`realtime/test/realtime/redis_test.exs` is new. It asserts `redis://` gets no socket options, `rediss://` asks for the match fun, and, importantly, that the returned fun actually accepts `*.upstash.io` for `example.upstash.io`. Asserting only on the shape would have passed with the wrong `:public_key` helper, which is the mistake this is guarding.

`mix format --check-formatted`, `mix compile` and `mix test` (29 passed) all clean.

## Summary by CodeRabbit

- **New Features**
  - Added optional CA bundle support for verified PostgreSQL TLS connections, including Amazon RDS.
  - Added improved hostname verification for secure Redis connections using `rediss://`.

- **Documentation**
  - Documented the `DATABASE_SSL_CA_FILE` setting, its default behavior, and the configuration required for Amazon RDS.
  - Updated split-deployment guidance for PostgreSQL certificate verification and the bundled AWS truststore.

### DIFF
--- a/deploy/docker/realtime.Dockerfile
+++ b/deploy/docker/realtime.Dockerfile
@@ -28,7 +28,7 @@ RUN mix release
 # the crypto NIF fails to load against an older libcrypto (missing symbols).
 FROM alpine:3.23
 
-RUN apk add --no-cache libstdc++ openssl ncurses-libs wget
+RUN apk add --no-cache libstdc++ openssl ncurses-libs
 RUN adduser -D -u 1000 warmbly
 
 # Amazon RDS presents a chain rooted in an RDS CA that is in no public trust
--- a/deploy/docker/realtime.Dockerfile
+++ b/deploy/docker/realtime.Dockerfile
@@ -28,8 +28,19 @@ RUN mix release
 # the crypto NIF fails to load against an older libcrypto (missing symbols).
 FROM alpine:3.23
 
-RUN apk add --no-cache libstdc++ openssl ncurses-libs
+RUN apk add --no-cache libstdc++ openssl ncurses-libs wget
 RUN adduser -D -u 1000 warmbly
+
+# Amazon RDS presents a chain rooted in an RDS CA that is in no public trust
+# store, and Postgrex verifies the peer against the system store by default.
+# Shipping AWS's truststore lets an operator opt in with
+# DATABASE_SSL_CA_FILE=/etc/ssl/rds/global-bundle.pem; nothing here changes the
+# default, because pointing every install at an RDS-only store would break a
+# Postgres fronted by a public CA.
+RUN mkdir -p /etc/ssl/rds && \
+    wget -qO /etc/ssl/rds/global-bundle.pem \
+      https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem && \
+    chmod 0644 /etc/ssl/rds/global-bundle.pem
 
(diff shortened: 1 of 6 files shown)


meszmate:
Good catch, fixed in 23ab7ba. A blank `DATABASE_SSL_CA_FILE` is now normalised to unset before the `cond`, so it falls through to the system-store default rather than passing `cacertfile: ~c""` with `verify_peer`.

This is the same rule the error-reporting section of this file already states for the DSN ("a variable that is present but blank counts as unset ... compose passes every optional variable through as `""`"), so the new code should have followed it from the start.

Verified by evaluating `config/runtime.exs` with `Config.Reader.read!/2` across all four cases, since this file is only read at release boot and nothing else exercises it:

| `DATABASE_SSL` | `DATABASE_SSL_CA_FILE` | resulting `:ssl` |
|---|---|---|
| `false` | any | `false` |
| `true` | unset | `true` (unchanged default) |
| `true` | `"   "` | `true` (this fix) |
| `true` | a path | `[verify: :verify_peer, cacertfile: ..., customize_hostname_check: ...]` |

PR #262, opened 2026-08-29
feat: self-hosted instances warm their mailboxes in the hosted pool (Warmbly Cloud link)

meszmate:
## What

A self-hosted Warmbly instance can link to a Warmbly Cloud workspace and have its mailboxes warmed in the hosted pool, while campaigns, contacts, inbox and storage stay on the instance.

- **Link**: device-code handshake. The instance shows a code (Settings > Warmbly Cloud), a member approves it at `app.warmbly.com/connect`, the instance receives a one-time `wpl_` token (sealed with `CREDENTIALS_ENCRYPTION_KEY`).
- **Enroll**: SMTP/IMAP mailboxes are created on the cloud as warmup-only accounts in the approving workspace: no history import, every non-warmup message is dropped before storage, warmup runs on the cloud's workers. Local warmup stands down for enrolled mailboxes.
- **Pricing**: free for 10 enrolled mailboxes per workspace, unlimited with any paid subscription (a `Self-hosted pool` $15 plan is seeded; attach its Stripe price in admin). `BILLING_PROVIDER=none` is unlimited.
- **Pool tiers**: paid is served first; when a tier has fewer than 25 eligible recipients, partner selection borrows the other tier's mailboxes that have been healthy members for at least 3 days.
- **UI**: animated three-step flow (link, pick mailboxes, done), steady-state table with health, today/7-day counts, pause/resume, disconnect; cloud-side "Linked instances" page; `/connect` approval page.
- **Docs**: `guides/warmbly-cloud.mdx`, configuration reference (`WARMBLY_CLOUD_URL`, `INSTANCE_NAME`, `WARMBLY_VERSION`), endpoints scope map, export/import note. Marketing copy: self-hosting is free, the pool link arrives with Warmbly Cloud.
- **Fix found while testing**: `emailRepository.SetWarmupLifecycle` re-read the row through the org-scoped `Get` with the user id, so every warmup start/pause/resume returned 404 after the write and skipped pool membership sync and the first task seed.

## Verified

`make lint`, web/docs typecheck+lint+build, orgtransfer coverage test, the repository prepare sweep, and a full end-to-end run with two isolated backends (cloud + self-hosted): link, approve, poll, enroll, pool membership, pause/resume, local warmup cancellation, unenroll, disconnect.

## Limits

- Google/Microsoft OAuth mailboxes cannot be enrolled (refresh grants are bound to the instance's own OAuth app); the UI explains and points to SMTP/IMAP.
[...]

PR #463, opened 2026-09-12
feat: fix the Warmbly Cloud pool link across both roles (#262)

meszmate:
## What

A deep pass over the Warmbly Cloud pool link shipped in #262, fixing every defect found in both halves of the feature (the self-hosted `cloudlink` side and the cloud `poollink` side), the worker and consumer paths, the routes and the docs.

## The defects

**An enrolled mailbox never left this instance's own warmup pool.** `IsEnrolled` stopped the local warmup *task*, but nothing removed the mailbox from `warmup_pool_participants`, so local mailboxes kept selecting it as a partner forever. That mail carries a local token; the consumer asks the cloud to vouch for it, the cloud says no, and it is filed as ordinary mail: warmup email in the customer's unibox, and the local sender's token never consumed. Enrolling now takes the mailbox out of the local pool (`canUseWarmupPool` plus a new `SyncWarmupPool`); unenroll and disconnect put it back.

**The cloud's own warmup mail landed in the unibox whenever the verify header did not survive delivery**, which the consumer's own comment says is every send from a Microsoft mailbox. Only the cloud holds the token, so it is the only side that can recognise the message. Added `POST /pool-link/instance/mailboxes/:remoteId/warmup-deliveries`, matching on `warmup_received` and on warmup tokens **ignoring `consumed_at`**: instance and cloud read the same mailbox, so whoever synced first must not decide what the other sees.

**`GET /api/v1/internal/cloud-link/token/:id` mints a live Gmail or Graph access token but sat behind `INTERNAL_API_TOKEN`**, the credential the internet-facing tracking and forms services also carry. Moved to the broker group (`NODE_BROKER_TOKEN`, falling back as the other two do); the worker presents that token.

**`SetLifecycle` never checked the mailbox's workspace**, so pause and resume acted on any account id across organizations.

**An unmapped status from the cloud was reported to the dashboard as a success.** `errx.Code(502)` is outside errx's table, `codeToHTTP` returns 0, and gin leaves that at 200, so a failed enroll looked like it worked. Fixed at the call site and hardened `errx.JSON` / `Handle` to answer 500 for any unknown code.

**An enrolled mailbox vanished from Settings > Warmbly Cloud once it went inactive** (the list query is active-only), leaving no way to see, pause or unenroll it while the cloud kept warming it.

**A failed local mirror row stranded the cloud mailbox.** It stayed linked to the instance, hidden from the adoptable list and refused with `pool_link_already_adopted` on retry, unreachable from either side. It is released now, as the sibling path already did.

**The one-time handshake could lose its token.** An instance with no `CREDENTIALS_ENCRYPTION_KEY` boots fine, so the link was approved on the cloud and then failed to store, stranding an instance nobody holds the token for. Refused up front with a message naming the variable.

**An approved code the instance never claimed kept a usable `wpl_` bearer token in plaintext** until some later sweep. `cli_auth_codes` already blanks these the moment the code expires; pool link did not.

Also: the stale workspace name on the flow's final step, `NODE_BROKER_TOKEN` threaded through the compose env anchor so backend and worker cannot disagree, and two dead methods removed (`poollink.IsLinkedMailbox`, `CountMailboxesForOrganization`).

## A deliberate call

Removing an enrolled mailbox from the local pool deletes its `warmup_pool_participants` row, which carries its local health record. That is the same mechanism and the same semantics the existing not-entitled path already uses, and it keeps membership as the single choke point every partner-selection query goes through; a selection-time predicate would preserve the record but has to be repeated in each query and is silently reintroducible. On a self-hosted instance the pool is the owner's own mailboxes, so nothing crosses a tenant boundary.

## Verified

- `make lint` (golangci-lint, gofmt, migrations, join / installer / split-cloud / dockerfile checks), `pnpm typecheck` and `pnpm lint` in `web/` and `docs/`, `go vet ./...`
- `TestLiveEveryQueryPrepares` against a freshly migrated database, so the new SQL is preparable
- New `internal/repository/pool_link_live_test.go`: the consumed-token case that motivates the new lookup, the `warmup_received` match, refusal of ordinary mail, and the expired code losing its plaintext token
- A backend booted on an isolated database and driven over HTTP: the shared internal token now gets 401 on the token route while the broker token reaches the handler, and the broker token does not widen access to the rest of the internal API; the public handshake start and poll; the new endpoint answering `valid:true` for the cloud's consumed-token mail, `valid:false` for a prospect, 404 for a mailbox the instance does not hold, and 401 unauthenticated

No migration.

## Summary by CodeRabbit

- **New Features**
  - Cloud-managed mailboxes can identify cloud warmup messages even when verification headers are missing.
  - Linked instances exclude cloud-managed mailboxes from local warmup handling.
  - Connection completion displays the linked workspace name.
  - Broker authentication supports a dedicated node token for mailbox access-token operations.

- **Bug Fixes**
  - Improved cleanup for failed enrollment and disconnected cloud links.
  - Error responses now return valid, consistent status codes.
  - Expired connection credentials are cleared promptly.

- **Documentation**
  - Updated token configuration and cloud warmup guidance.

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -109,8 +109,14 @@ func main() {
 		log.Fatal(err)
 	}
 	// Mailboxes managed by Warmbly Cloud send with access tokens the backend
-	// brokers; the refresh grant never reaches the worker.
-	tokenBroker, err := repository.NewHTTPBrokeredTokenClient(internalBaseURL, internalToken)
+	// brokers; the refresh grant never reaches the worker. Minting one is a
+	// broker operation, so it carries NODE_BROKER_TOKEN where the instance
+	// issues a separate one and the shared internal token otherwise.
+	brokerToken := os.Getenv("NODE_BROKER_TOKEN")
+	if brokerToken == "" {
+		brokerToken = internalToken
+	}
+	tokenBroker, err := repository.NewHTTPBrokeredTokenClient(internalBaseURL, brokerToken)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/deploy/split-cloud/control-plane.env.example
+++ b/deploy/split-cloud/control-plane.env.example
@@ -39,10 +39,11 @@ CREDENTIALS_ENCRYPTION_KEY=
 INTERNAL_API_TOKEN=
 SECRET_KEY_BASE=
 
-# The credential for the two routes that open a sealed key and sign a blob
-# operation. Falls back to INTERNAL_API_TOKEN when unset; set it to its own
-# value here, because the tracking and forms services are internet-facing and
-# carry the shared one. Nodes receive this, the edge services do not.
+# The credential for the routes that open a sealed key, sign a blob operation
+# and mint a managed mailbox's access token. Falls back to INTERNAL_API_TOKEN
+# when unset; set it to its own value here, because the tracking and forms
+# services are internet-facing and carry the shared one. Nodes receive this,
+# the edge services do not.
 NODE_BROKER_TOKEN=
 
 # --- storage ----------------------------------------------------------------
--- a/deploy/split-cloud/node/worker.env.example
+++ b/deploy/split-cloud/node/worker.env.example
@@ -24,7 +24,8 @@ ENCRYPTED_KEYS_PROVIDER=http
 ENCRYPTED_KEYS_BACKEND_URL=https://api.example.com
 ENCRYPTED_KEYS_WORKER_TOKEN=
 INTERNAL_API_TOKEN=
-# Only when the control plane sets one. Used for the brokered providers below.
+# Only when the control plane sets one. Used for the brokered providers below
+# and for the access token a mailbox Warmbly Cloud manages sends with.
 NODE_BROKER_TOKEN=
 
 # --- the bus and cache ------------------------------------------------------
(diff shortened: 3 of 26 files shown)


meszmate:
Addressed all four review threads.

Three taken as reported: `IsWarmupDelivery` now refuses the sender-and-subject fallback when the message and the token both carry a Message-ID and they differ (provably a different message, and matching on subject alone would drop the owner's real mail), with a live regression test that was mutation-tested by removing the guard and watching it fail; disconnect aborts when the enrolled mailboxes cannot be enumerated instead of deleting the link row that is the only way to retry; and the configuration reference now says NODE_BROKER_TOKEN has to hold the same value on the control plane and on every node.

The fourth I took differently. Returning false from the cloud-enrollment lookup on error does not fail closed, it fails *open in the other direction*: `syncWarmupPoolMembership` treats false as "remove from all pools", which deletes the `warmup_pool_participants` row and with it the mailbox's health record, so a transient database blip during any mailbox update would evict a healthy mailbox. The question has three answers, not two, and this repo already has a rule for the third one ("Only a definite 'not entitled' evicts" in the warmup reconciler, and the matching switch in `email_task.go`). So the check moved out of the `canUseWarmupPool` bool and into `syncWarmupPoolMembership`, where enrolled removes, not-enrolled proceeds, and an unreadable answer leaves the membership exactly as it is.

PR #464, opened 2026-09-12
fix: stop the contacts and campaign-leads table overflowing its panel on a long company name (#461)

### LINKED ISSUE #461: Bug: Horizontal Scrollbar Appears After Loading More Contacts in campaign
When I click the “Load More Contacts” button, additional contacts are loaded successfully, but an unwanted horizontal scrollbar appears. The contact list should remain within the available width without triggering horizontal scrolling.

campaign > Leads - loadmore button in bottom

Image

In the company column, the company name is very long. I think it is happening due to content. 

If content should be long, then it should be text truncated, and one hover should show it in a tooltip full text

@meszmate

meszmate:
Closes #461.

## The bug

`ContactsTable` (shared by `/app/contacts` and the campaign **Leads** tab) rendered `<table className="w-full text-left">`, i.e. **auto** table layout. The Company cell put `truncate` on the `<td>` itself, which sets `white-space: nowrap` but gives the cell no width constraint. In auto layout a cell's *min-content* width is then the full text width, so one long company name forced the whole table wider than the content panel. Loading more contacts simply brought in the row that had one. `text-overflow: ellipsis` on a `<td>` never rendered either, so nothing truncated.

The embedded Leads table has no `PageBody` wrapper, so that overflow propagated up to the app shell's scroll container — the full-width scrollbar in the report.

Name, Phone and Sender had the same unbounded shape; Company was just the first to hit it.

## The fix

**`table-fixed`.** Column widths now come from the header row alone, so cell content can only ever truncate, never widen the table. Every column carries a width and every cell clips — content overflowing a fixed column still extends the scroll container, so the clipping is load-bearing, not cosmetic. `Name` is deliberately the *only* auto-width column, so it absorbs all the slack rather than splitting it with Company and Phone.

**Truncate + tooltip on hover, only where clipped.** New shared `web/src/lib/helper/titleWhenClipped.ts` exports a `clippedTitle` props object: on mouseenter it attaches the full text as a native `title` when `scrollWidth > clientWidth`, and takes it off again on mouseleave so nothing stale outlives the hover. A column that fits stays quiet. Native `title` matches this table's existing convention (current step, sender, categories already use it). Applied to name, email, company, phone, category chips and the status pill labels — but not where a cell already carries a more useful title (a lead's failure reason wins over the word "Failed").

**Column staging retuned.** Fixed layout exposed a second, pre-existing overflow: the Leads table's fixed columns alone summed to ~612px at `md`, where the panel is only ~512px (viewport − the `w-64` sidebar), so it overflowed there regardless of company names. Note the panel *narrows* from `sm` to `md` because the sidebar appears, which the old `md:` reveals had backwards. Columns are now staged so the sized widths never sum past the panel at any breakpoint, keeping the auto Name column ≥168px everywhere:

| Column | Contacts | Leads |
|---|---|---|
| Company | `lg` (was `md`), `w-36`/`w-40` | `xl` (was `md`) |
| Phone | `xl` (was `lg`), `w-36` | not shown (it is in the contact drawer) |
| Opened / Clicked / Replied | — | `lg` (was `md`), `w-24` |
| Current step | — | `xl` (was `md`), `w-32` |
| Sender | — | `2xl` (was `xl`), `w-36` |
| Campaigns | `lg` (was `md`), `w-28` | — |
| Added / Last activity | `md`, `w-24` | `2xl` (was `md`), `w-32` |
| Status / Progress | `w-12` below `sm` (the pill is its icon alone there; the label goes to `sr-only` so the column keeps its accessible name), `w-32` from `sm` | same |

Worked example at the reporter's 1657px: panel 1401px, sized columns 1096px, leaving ~305px for Name — truncated with a tooltip, no scrollbar. Narrowest case (360px phone) leaves Name 192px.

**Category chips.** With a fixed-width Name column the chip group could no longer be `shrink-0` — two long tags took the whole cell and left no name. It is capped at 45% of the name line and both sides ellipsize into their tooltip instead. `CategoryChip` gained an opt-in `clamp` prop for this; `compact` on its own still means "dense", so unibox thread labels are unchanged.

## Verification

`pnpm typecheck` clean, `pnpm lint` 0 errors (85 pre-existing warnings, count unchanged), `vitest run` 231/231 across 26 files including 5 new ones for the tooltip helper.

Not verified in a browser — worth an eyeball on the Leads tab around 1280 and 1536, where the column staging changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved horizontal scrolling issues in the contacts table.
  - Improved responsive column visibility across contact and campaign lead views.

- **Improvements**
  - Long contact details, statuses, and category labels now truncate cleanly while providing full-text tooltips.
  - Category tags use a compact display with a count for additional categories.
  - Updated table column sizing improves readability and prevents content from disrupting the layout.


meszmate:
Addressed the accessible-name finding in 8a442b44.

Both pills now carry their word in a permanently `sr-only` span with the visible copy marked `aria-hidden`, rather than toggling one element with `sm:not-sr-only`. `not-sr-only` resets `overflow` and `white-space`, which is exactly what the `truncate` on that label depends on at `sm+`, so the one-element version would have traded a narrow-screen a11y gap for a wide-screen clipping regression. Same treatment applied to the column header, which had the same gap.

Not acting on the docstring-coverage check: this repo's conventions ask for short single-line comments stating the non-obvious constraint, not JSDoc on every component helper, and most of the counted functions are ones this diff only touched a line in.

PR #465, opened 2026-09-12
fix: mailbox onboarding accepts connections that cannot work, and hides why

meszmate:
Connecting a Gmail mailbox on production fails with `400 Bad Request: Could not read account details from the provider.` and that is the entire record. Nothing in the logs, nothing in error tracking, no way to tell what went wrong. Chasing that turned up two more ways a mailbox can be accepted and then quietly not work.

## 1. The reason the provider refused was discarded

The OAuth flow is fine — the callback returns 200 with every scope granted. The next call fails, and `fetchGmailOwner` reads the status and the body and throws both away:

```go
body, _ := io.ReadAll(resp.Body)
if resp.StatusCode != http.StatusOK { return nil, errx.ErrEmailOnboardUserInfo }
```

Five distinct failures collapsed into one sentence: the Gmail API was never enabled on the project (403, and by far the most common), the token was revoked (401), the provider is down, the body did not parse, the body named no address. The first is a console toggle, the second is the user's problem, the third is nobody's, and they were indistinguishable. Because a handled 400 raises no exception, error tracking never saw any of it either.

`ownerLookupFailed` now records provider, stage, status and a truncated body, reports it with the request scope attached, and logs at warn. Both the Gmail and Outlook fetchers route every failure through it. The user-facing message is unchanged on purpose: the cause is ours, and "the Gmail API is not enabled in project 1010273043313" is not a sentence to show a customer.

## 2. A partly granted consent was accepted

Google's consent screen has a checkbox per permission. Untick "send email on your behalf" and the flow still completes and still returns a token. `OAuthFinish` went straight from `cfg.Exchange` to storing the mailbox without ever looking at what was granted, so the mailbox connected cleanly and then failed on its first campaign send, days later, with a provider 403 nobody could trace back to a checkbox.

`checkGrantedScopes` compares the token's `scope` against what was requested and refuses the connection naming the missing permission in the consent screen's own words ("Missing: send email on your behalf").

Two things it deliberately does not do:

- **It does not treat a broader grant as a denial.** Gmail scopes nest: `gmail.modify` confers reading and sending, `gmail.compose` confers sending. Someone who granted the broad permission but not the narrow one has the capability and is let through.
- **It does not treat an absent `scope` field as a denial.** A provider that returns no scope has told us nothing, which is not the same as granting nothing. That case is reported rather than guessed at, so a provider that stops returning it cannot turn this check into a silent no-op.

## 3. A first connect with no refresh token was stored

That mailbox has about an hour to live: the access token expires and nothing can renew it. `reauth.go` already tolerates an absent refresh token because it keeps the stored one, but on a first connect there is nothing to fall back on. Now refused, with an error telling the user to revoke access and reconnect.

## Tests

`onboarding_owner_test.go` drives the package HTTP client at a local server and covers the four Gmail failure shapes, the success path and body truncation; then partial consent for Gmail and Outlook, full consent, the broader-scope-satisfies-narrower case, and the absent-scope-field case.

`make lint` and `go test ./internal/app/email/` pass.

### DIFF
--- a/internal/app/email/onboarding.go
+++ b/internal/app/email/onboarding.go
@@ -132,6 +132,15 @@ func (s *emailService) OAuthFinish(ctx context.Context, userID, code, state stri
 		return nil, false, errx.ErrEmailOnboardExchange
 	}
 
+	// A consent screen lets the person untick individual permissions and still
+	// returns a token, so this is the last point at which a half-granted
+	// mailbox can be refused. Storing one instead means a mailbox that looks
+	// connected and fails on its first send, days later, with a provider 403
+	// nobody can trace back to a checkbox.
+	if xerr := checkGrantedScopes(ctx, provider, cfg.Scopes, tok); xerr != nil {
+		return nil, false, xerr
+	}
+
 	owner, xerr := fetchInboxOwner(ctx, provider, tok.AccessToken)
 	if xerr != nil {
 		return nil, false, xerr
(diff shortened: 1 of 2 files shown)


meszmate:
All three addressed in 8db57a9.

**PII in logs — you were right, and it contradicted a claim in my own PR description.** I wrote that the body "is only ever read on a failure response." That is false for three of the five stages: Gmail `decode`, Gmail `no_address` and Outlook `decode` all carry status 200, so the body is a *successful* profile payload — the mailbox owner's address, and for Outlook their display name too.

Rather than pass `nil` at those three call sites, the decision moved into `diagnosticBody(status, body)`, which returns nothing for any 2xx. Doing it at the call sites would have left the next person adding a failure stage free to get it wrong; doing it in one place means they cannot.

**Bounded reads:** both providers now read through `io.LimitReader(resp.Body, maxProviderBody)` (64 KiB) before anything looks at the body.

**Asserting the diagnostics:** rather than capture zerolog and stand up an httptest server for PostHog — brittle, and it would test the SDKs more than this code — the reportable-content decision is now a pure function with direct tests:

- `TestDiagnosticBodyNeverRecordsASuccessfulPayload` — a real profile payload against 200/201/204/299, asserting nothing is recorded. This is the regression test for the issue you raised.
- `TestDiagnosticBodyKeepsFailurePayloads` — a 403 still yields the "Gmail API has not been used in project…" text, so the fix does not defeat the purpose.
- `TestDiagnosticBodyTruncatesToTheLimit` — exactly the 512-byte prefix plus the marker.
- `TestProviderBodyReadIsBounded` — a response 4x the cap.

`make lint` and `go test ./internal/app/email/` pass.

PR #466, opened 2026-09-12
feat: AI is a paid feature, so take it off the free plan

meszmate:
The free plan is the connect-and-see-it-work tier: mailboxes, sync, a couple of small campaigns. It was also granting 50 AI credits.

Zeroing the allowance alone would have made things worse, not better, so this is three changes that only make sense together.

## Why the gate had to move too

`CanUseInboxAgent` was already paid-only. `CanUseWritingAssistant` was not:

```go
return sub.HasPaidSubscription() || sub.IsInFreeTrial(), nil
```

So a trial workspace was *shown* the assistant, and the 50 credits existed to make that work. Drop the credits and leave the gate, and a free user gets a visible button that can only ever answer "you have no credits" — a broken feature rather than a locked one, and the worst of both outcomes.

The assistant now gates exactly like the inbox agent. A test asserts the two agree, so they cannot drift apart again.

## The three parts

- **Migration `000148`** sets the free plan's `monthly_credits` to `0`. The trial grant in `internal/app/trial` only fires when `monthly_credits > 0`, so it stops by itself with no code change there.
- **`internal/seed/plans.go`** matches, so a fresh install and a migrated one agree.
- **`CanUseWritingAssistant`** becomes paid-only.

Existing balances are deliberately left alone. Anyone already granted the allowance keeps it; clawing back credits from an account that was given them is a worse surprise than the inconsistency, and the down migration restores the plan value rather than re-granting.

## Docs

`ai-credits.mdx` said "Free trial | 50 (granted once, at trial start)". It now says AI is paid-only, and the intro says so too rather than leaving it to a table row.

That same table also named plans that do not exist — `Pro` and `Enterprise`, where the real ones are `Grow` and `Business`. Fixed while here, since a pricing table naming the wrong plans is its own bug.

`ai-assistant.mdx` now leads with the paid requirement before the role permission, because the permission is irrelevant if the plan locks it.

## Note on a nearby trap

`plans.ai_generation` looks like it gates this and does not. It is written and read by `pg_plan.go` and carried on the model, but nothing consults it — the free plan has had `ai_generation = false` the whole time while the assistant was available. Left alone here rather than widening the change, but it should either gate something or go.

`make lint` (including `make check-migrations`, 148 versions, no duplicates) and `go test ./internal/app/feature/` pass.

## Summary by CodeRabbit

* **Changes**
  * AI features now require an active paid plan.
  * Free trials no longer include AI credits, and AI features remain locked until upgrading.
  * Updated plan documentation to show free-trial availability and renamed credit plans to Grow and Business.
  * Self-hosted installations continue to have unrestricted AI access.
* **Bug Fixes**
  * Aligned AI feature access checks for consistent paid-plan enforcement.

### DIFF
--- a/docs/content/docs/guides/ai-assistant.mdx
+++ b/docs/content/docs/guides/ai-assistant.mdx
@@ -11,7 +11,7 @@ Open it from the spark button or `Cmd/Ctrl + I`. The panel follows you between p
 The assistant acts as you: only what your role allows, only in your workspace, never another organization's data. Every change asks first, and **sending is never auto-approved**, even with "Always allow" on.
 </Callout>
 
-Access is the **Use AI** role permission (on for admins and managers, off for viewers); without it the assistant and AI drafting are hidden. Admins can cap per-member spend by day, week, or month. See [AI credits](/guides/ai-credits/) and [Team and roles](/guides/team-roles/).
+The assistant needs a paid plan; a free trial has no credit allowance and the AI surfaces stay locked. Beyond that, access is the **Use AI** role permission (on for admins and managers, off for viewers); without it the assistant and AI drafting are hidden. Admins can cap per-member spend by day, week, or month. See [AI credits](/guides/ai-credits/) and [Team and roles](/guides/team-roles/).
 
 ## Conversations and the panel
 
--- a/docs/content/docs/guides/ai-credits.mdx
+++ b/docs/content/docs/guides/ai-credits.mdx
@@ -3,7 +3,7 @@ title: AI credits
 description: How AI credits work in Warmbly, what each AI action costs, the monthly allowance and reset rules, and how to buy top-up packs.
 ---
 
-AI features run on credits. Your plan grants a monthly allowance that refreshes each billing cycle, and top-up packs never expire.
+AI features run on credits, and they are paid-only: a free trial carries no allowance and the AI surfaces stay locked until the workspace is on a paid plan. Your plan grants a monthly allowance that refreshes each billing cycle, and top-up packs never expire.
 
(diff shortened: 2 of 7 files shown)


meszmate:
The asymmetry is real and I've documented it rather than papered over it (a618d06). Reasoning, because I did not take the provenance-tracking route you suggested:

The exposure is narrower than it first looks. The statement is already scoped to the single free-plan id, and only fires when the value is still the zero `up` would have left. What it cannot distinguish is a zero it wrote from a zero that was already there — you're right about that, and there is no marker to tell them apart.

Recording the prior value would mean either a leftover `_migration_000148_prior` table (which then has to be reasoned about for `orgtransfer/spec.go`, since anything not in `Tables` or `ExcludedTables` is the bug that rule exists to prevent) or migration bookkeeping in `admin_settings`, a product table. For one row of plan configuration, in a path the deployment guide explicitly says is forward-only, both felt like the more expensive mistake.

So I followed what this repo already does for data migrations that cannot fully reverse. `000147` lists three things that do not survive its round trip and says why that is acceptable; `000145` explains that its rows are recoverable *because* the up migration stamped them with a marker nothing else writes. `000148` now states its one case plainly:

> One case does not round-trip: a free plan that was ALREADY at zero before 000148 ran comes back as 50, because the up migration recorded no prior value to restore.

Blast radius is one configuration row an operator can edit in the admin panel, so the worst outcome is a rollback followed by setting it back. Happy to add the provenance table if you think that trade is wrong — but I did not want to introduce an unregistered table quietly.

`make check-migrations`: 148 versions, no duplicates.

PR #467, opened 2026-09-12
feat: plans an operator can grant, visible in the admin panel

meszmate:
There was no way to give any workspace a paid plan without Stripe. `HasPaidSubscription()` was:

```go
return s.StripeSubscriptionID != nil && s.Status.IsActive()
```

So an internal workspace, a design partner or a support gesture had exactly one route: hand-write a fake `stripe_subscription_id` into the table and hope no webhook ever reconciled against an id Stripe has never heard of. I checked for an existing concept — `comped`, `manual_plan`, `is_internal`, `granted_plan` — and there was none. Limit overrides move numeric caps only, never feature gates.

This is felt today: the instance owner's own workspace cannot send campaigns, use the unibox, or use AI, so the product cannot be evaluated from the account that runs it.

## The grant

`subscriptions` gains `managed_at`, `managed_by`, `managed_reason`, `managed_until`, mirroring `organizations.risk_override_*` — who, why, when, so a grant is answerable months later. A `CHECK` refuses a grant with no reason, because an unexplained paid workspace is the thing this is meant to prevent.

`managed_until` NULL is open-ended; a date lapses on its own, so a time-boxed grant needs nobody to remember to revoke it.

`IsManaged()` is checked **first** in `HasPaidSubscription()`, so a granted workspace is paid everywhere entitlements are read, and a workspace that later subscribes for real is unaffected either way.

`ManagedExpired()` is deliberately separate from `IsManaged()`. A lapsed grant is not the same as never having had one: the workspace drops back to free, but the reason stays readable instead of the change looking like it happened for no reason.

## Two things that would have bitten

- **`SetManagedPlan` is off `Update`.** Subscription writes come from Stripe webhooks, and a routine write must not be able to clear a grant as a side effect.
- **Both subscription read paths carry the columns.** `subscriptionColumns` feeds the entitlement path; `GetWithLimits` feeds the customer's own billing page. Updating only the first would have made a granted workspace paid everywhere except where the customer looks.

## Admin panel

- **Badge** on the organization list: `managed`, or `grant lapsed`, with the reason on hover. Included in the CSV export.
- **Filter**: "Managed plan", alongside the existing Enterprise and overrides toggles, so "who is paid because we said so" is one click.
- **Card** on the workspace, under Plan: the grant with its reason, who made it and when it expires, plus grant and revoke. Reading needs `view_organizations`; both writes need `manage_organizations`, so an admin without it sees the record rather than buttons that can only answer 403.
- Both writes go through `LogAdminAction`, so the grant is on the platform-admin trail.

`GET /admin/plans` is new because the customer endpoint returns public plans only, and a grant is usually onto a private one.

## Guarded

A grant dated in the past is refused — it would read as "granted" in the audit trail while entitling nothing. An unknown plan id is a 400 naming the problem rather than a foreign-key violation, which is why the org service now takes a `planRepo`.

## Tests

`subscriptions_managed_test.go` covers open-ended and time-boxed grants, a lapsed one, never-granted, that `ManagedExpired` distinguishes lapsed from never, that a real Stripe subscription is unaffected in both directions, and nil receivers.

`make lint` (149 migrations, no duplicates), `go test ./internal/models/`, `pnpm typecheck` and `pnpm lint` in `admin/` all pass.

## Summary by CodeRabbit

* **New Features**
  * Administrators can grant, update, view, and revoke managed plans for workspaces without requiring Stripe billing.
  * Managed plans support reasons, optional expiry dates, status badges, and plan entitlement details.
  * Organization lists can filter by managed-plan status and display lapsed grants.
  * CSV exports now include managed-plan status.

* **Documentation**
  * Added billing guidance explaining managed plans, expiry, and revocation behavior.

### DIFF
--- a/admin/src/app/dashboard/OrganizationDetailPage.tsx
+++ b/admin/src/app/dashboard/OrganizationDetailPage.tsx
@@ -55,6 +55,7 @@ import type {
 } from "@/lib/api/models/admin";
 import { OrganizationOverridesDialog } from "./OrganizationOverridesDialog";
 import { OrganizationRiskCard, RiskBadge } from "./OrganizationRiskCard";
+import { OrganizationManagedPlanCard } from "./OrganizationManagedPlanCard";
 
 const TABS = [
     { id: "overview", label: "Overview", icon: LayoutDashboard },
@@ -192,6 +193,11 @@ export default function OrganizationDetailPage() {
                     )}
                 </SummaryCard>
             </div>
+
+            <section className="mt-6">
+                <h2 className="text-sm font-semibold mb-2">Plan</h2>
+                <OrganizationManagedPlanCard orgId={org.id} />
+            </section>
 
             <section className="mt-6">
                 <h2 className="text-sm font-semibold mb-2">Abuse posture</h2>
--- a/admin/src/app/dashboard/OrganizationManagedPlanCard.tsx
+++ b/admin/src/app/dashboard/OrganizationManagedPlanCard.tsx
@@ -24,6 +24,15 @@ import {
 } from "@/lib/api/client/admin/organizations";
 import type { ManagedPlan } from "@/lib/api/models/admin";
 
+/** A date input gives "YYYY-MM-DD" with no time. Reading that with `new Date`
+ *  yields UTC midnight, so a grant made for today is already expired for
+ *  anyone west of UTC, and the date displayed back can be the previous day.
+ *  The grant runs to the end of the chosen day in the operator's own zone. */
+function endOfLocalDay(day: string): string {
+    const [y, m, d] = day.split("-").map(Number);
+    return new Date(y, m - 1, d, 23, 59, 59, 999).toISOString();
+}
+
 function fmt(ts?: string | null) {
     if (!ts) return null;
     return new Date(ts).toLocaleDateString(undefined, { year: "numeric", month: "short", day: "numeric" });
@@ -85,7 +94,7 @@ export function OrganizationManagedPlanCard({ orgId }: { orgId: string }) {
             grantOrganizationManagedPlan(orgId, {
                 plan_id: planId,
                 reason: reason.trim(),
-                until: until ? new Date(until).toISOString() : null,
+                until: until ? endOfLocalDay(until) : null,
             }),
         onSuccess: (m) => {
             applied("Plan granted")(m);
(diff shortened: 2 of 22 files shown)


meszmate:
All four addressed in f1e22c2. The plan-overwrite one was a real bug I introduced, so thank you for it.

**Store the managed plan separately — fixed as suggested, not worked around.** `SetManagedPlan` was overwriting the shared `plan_id`, so a workspace paying Stripe for plan A and granted plan B would have been stranded on B after revocation: paying for one thing, entitled to another, permanently.

There is now a `managed_plan_id` column beside `plan_id`, never on top of it, and `Subscription.EffectivePlanID()` resolves which one decides entitlements. I considered just refusing a grant over an active Stripe subscription, which would have been a one-line guard, but that removes a legitimate case (a temporary upgrade for an existing customer) to avoid doing the work.

The nine `GetByID(ctx, sub.PlanID)` call sites split deliberately:

- **Five moved to `EffectivePlanID()`** — the entitlement readers: `contact/handler.go`, both in `feature/gate.go`, `subscription/service.go`, `worker/assignment.go`.
- **Four stayed on `PlanID`** — all in `stripe/service.go`. Those reason about what the customer *pays*: checkout comparison, subscription created/updated, invoice paid. Resolving a grant there would have Stripe prorating against a plan the customer never bought.

That split surfaced a limitation worth stating rather than hiding, so it is now in the billing guide: monthly AI credits are granted on `invoice.paid`, and a granted plan raises no invoice, so a managed workspace gets no recurring credit allowance.

**Repository error vs missing plan** — `perr != nil` is now `errx.Internal` with the error captured; only `plan == nil` is `BadRequest`.

**Local end-of-day expiry** — right, and worse than you described: with UTC midnight, an operator west of UTC granting "until today" created an already-expired grant, which the service would then have refused outright. `endOfLocalDay` runs it to 23:59:59.999 in the operator's own zone.

**The index** — I removed it rather than making it concurrent. It only ever served the admin list filter, a rare query on a table with one row per workspace, and the planner would likely drive that join from `organizations` anyway. It was speculative, so the honest fix was to delete it, which also retires the transaction-wrapping question entirely.

New tests cover a lapsed grant falling back to the paid plan, that `PlanID` is never overwritten, and a grant recorded with no plan id not shadowing the real one.

`make lint` (149 migrations, no duplicates), `go test ./internal/models/ ./internal/app/feature/`, and `pnpm typecheck` + `pnpm lint` in `admin/` all pass.

PR #468, opened 2026-09-12
A mailbox re-reading its own mail is not a forged warmup token

joaoppa:
## The problem

`handleWarmupEmail` records an invalid-token attempt on three paths, and none of them asks whether the message could be the mailbox's own. On a live self-hosted instance that made every single attempt a false positive.

```
invalid attempts on file                                211
  re-read of a token that had expired                    88   41.7%
  the Sent copy, carrying the recipient's token          67   31.8%
  reconnected mailbox, token cascaded away               56   26.5%
  a token belonging to neither side of the pair           0      0%
```

The last row is the one the detector exists to catch. It is empty, and three attempts in 24h blocks a mailbox from the pool for 30 days.

### The three shapes

**Expired re-read.** `GetWarmupToken` filters `consumed_at IS NULL AND expires_at > NOW()`, so a token that is on file but past its 7-day window comes back nil and the mailbox is charged the `scoreDelta: 5` path — the most punishing of the three. All 88 of these were expired, none consumed; an IMAP resync re-reading old mail is enough. 88 × 5 is 440 spam score against a block threshold of 50.

**Sent copy.** Since Sent folders sync, the sender's own copy of every warmup send reaches the consumer under the *sender's* account still carrying the token, and the recipient check fails. Confirmed by joining on `sender_account_id`, which is how these separate from a genuine stranger's token.

**Reconnect.** `warmup_tokens` has `ON DELETE CASCADE` on both account foreign keys, and reconnecting a mailbox writes a new row with a new id. The cascade takes every token the old row owned; the new row syncs the same Inbox and presents tokens that no longer resolve. On this instance all 56 came from one mailbox in one burst:

```
account row created   02/09 00:18:44
first attempt         02/09 00:18:49
last attempt          02/09 00:18:52
```

Five seconds. Reconnecting a mailbox — routine, and the documented fix for an expired OAuth grant — can block that mailbox within the minute. The same mechanism leaves orphan rows in `warmup_received` (114 here).

## The change

A token that cannot be verified is recorded only when it could not be the mailbox's own mail. The decision is a pure function so it can be exercised without a mailbox or a token store:

```go
func warmupTokenIsOwnMail(folder string, accountID uuid.UUID, tok *models.WarmupToken, mailboxAge time.Duration) bool
```

- filed in Sent → own mail
- the row still exists with `GetWarmupToken`'s filters removed (`FindWarmupToken`, already in the repository) and this account is its sender or recipient → a re-read
- the mailbox row is younger than `WarmupReconnectGraceMinutes` (60) → first sync after a reconnect
- anything else is still recorded, on the same path and with the same score as before

Nothing is relaxed for a token that belongs to another pair, which is the only shape that was ever evidence.

## Tests

`TestWarmupTokenIsOwnMail` covers the three exonerating shapes and three that must stay signal (a stranger's token, an unknown token on an established mailbox, an unknown token with no mailbox age). Verified both ways: on the previous behaviour the four ownership cases fail.

```
--- FAIL: TestWarmupTokenIsOwnMail/sent_copy_carries_the_recipient's_token
--- FAIL: TestWarmupTokenIsOwnMail/expired_token_of_mail_this_mailbox_received
--- FAIL: TestWarmupTokenIsOwnMail/consumed_token_of_mail_this_mailbox_sent
--- FAIL: TestWarmupTokenIsOwnMail/first_sync_after_a_reconnect,_token_cascaded_away
```

gofmt, `go build ./...`, `go vet`, `make check-migrations` and golangci-lint (v1.64.8) clean; package tests pass with `-race -count=2`.

## Not addressed here

The cascade itself. Losing a mailbox's warmup history on reconnect also costs its pool statistics and leaves the orphan `warmup_received` rows above; whether those keys should be `ON DELETE SET NULL`, and what that means for the pool health a reconnected mailbox inherits, is a larger question than this fix. The grace window stops the detector acting on it, and no more.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved warmup email validation to avoid incorrectly flagging a mailbox’s own sent, expired, or consumed messages as suspicious.
  - Reduced false alerts during reconnection by allowing a 60-minute grace period for recently created mailboxes when token details cannot yet be verified.
  - Continued identifying genuinely unknown or mismatched tokens as suspicious.
  - Improved handling of temporary token lookup failures so valid messages are not incorrectly flagged.


meszmate:
Merging this. The diagnosis is the best part: I checked each of the three shapes against the code and they all hold up. `GetWarmupToken` really does filter `consumed_at IS NULL AND expires_at > NOW()`, so the most innocent cause was taking the most punishing branch; the worker canonicalises Sent folders to `models.FolderSent`, so the Sent copies really do arrive here; and `FindWarmupToken` was already sitting unfiltered in the repository. Splitting the decision into a pure function was the right call, and testing the cases that must stay signal rather than only the ones being fixed is why the two defects below were findable at all.

I pushed three fixes rather than sending it back, since it is fixing a live instance. All three are in the same decision function.

**1. The window was reachable by a token that resolves to another pair.** The ordering put the age check last, so `tok != nil` with neither side matching fell through to it and a mailbox younger than 60 minutes was forgiven. That is the one row your own table calls the only shape that was ever evidence, and reconnecting resets the clock, so a mailbox that keeps reconnecting could probe the pool indefinitely without ever being charged an attempt. The token now decides whenever it resolves at all, and the window only sees the case where nothing resolves.

**2. A harvested token laundered through Sent.** `folder == FolderSent` returned true before the token was consulted, and Sent is writable over IMAP APPEND. The genuine Sent copy never needed that branch: its token names this mailbox as sender, so the `tok != nil` branch already covers it. Worth noting the test was modelling this one wrong, using `strangers` where a real Sent copy carries `sentByMe`, which is what hid it. Sent now only helps when the token has vanished entirely.

**3. A failed lookup was charged as a forgery.** `if t, err := FindWarmupToken(...); err == nil` treated a DB error and a missing row identically. `scanWarmupToken` returns `(nil, nil)` for no rows, so a non-nil error is only ever a failed question, and failing open there cannot weaken the detector. It matters because the blip arrives during exactly the re-read burst that makes three attempts. The mismatch call site also now passes the token it already holds, so that path does no second read and cannot be downgraded by one.

Every assertion is mutation tested the way you did it: each guard reintroduced, each failing exactly its own row and nothing else. `make lint` clean, `-race -count=2` clean.

**Two things I am not asking you to fix**, both filed:

- #476 for the cascade. You were right to leave it, and it is worse than the PR says: when the *partner* mailbox is re-added, the established mailbox that did not reconnect re-reads its own inbox, finds nothing, and is charged with no window to protect it, since its row is old. Plus the 114 orphan `warmup_received` rows you found.
- #477 for the window itself. It is outrun by the backfill, which the sync governor paces at 240/min against a cap reaching 100k messages, so a large history import is charged for everything after minute 60. And `email_accounts.created_at` is `timestamp without time zone`, so `time.Since` on it is wrong by the instance's UTC offset and the window silently does not apply at all on a non-UTC database. Both fail toward charging rather than toward a miss, which is why they did not hold this up. Your `InternalDate` is already on the message and makes the window unnecessary.

Thanks for this one. The evidence table in the description is what made it reviewable.

PR #474, opened 2026-09-13
fix: a granted plan unlocks nothing, and Turnstile never renders

meszmate:
Two production bugs found from a screenshot of the dashboard.

## 1. The granted plan was recognised everywhere except where it matters

The header showed `BUSINESS` and `50/25k` credits — both correct, both resolved through `EffectivePlanID()`. Meanwhile every sidebar entry was locked behind `STARTER` and the mailbox panel said *"1 of 10 free mailboxes used"*.

The client decides paid from Stripe's status alone:

```ts
const subSaysPaid = status === "active" || status === "trialing";
```

A granted plan never touches Stripe, so `status` keeps whatever it had — `incomplete` on a workspace that never subscribed. The workspace was fully entitled server-side and locked client-side, which is the most confusing shape this bug could take: the plan pill and the credit balance both said Business while nothing opened.

`SubscriptionWithLimits` now carries `managed`, resolved from `IsManaged()` so expiry is applied server-side and no client reimplements it. `useFeatureAccess` includes it in `isPaid`.

While there: `GetWithLimits` never loaded the plan at all, so `sub.data.plan.name` was always undefined and the dashboard fell through to whatever the org row said. It now loads the effective plan, which is what makes the pill correct rather than coincidentally correct.

## 2. Turnstile has not been rendering

From the production console:

```
[Cloudflare Turnstile] Invalid value for parameter "size",
expected "compact", "flexible", or "normal", got "invisible"
```

Cloudflare removed the `invisible` size. Three call sites still pass it: the login page and both `TurnstileModal`s (web and admin). The widget refuses to render, so no token is ever produced — on a deployment with `CAPTCHA_PROVIDER=turnstile` that is the login path.

`react-turnstile`'s typings still list `"invisible"`, which is why TypeScript never caught it and only the browser did.

Replaced with the supported equivalent for programmatic use: `execution="execute"` defers the challenge until `.execute()` is called, which is what this code already does, and `appearance="interaction-only"` keeps it out of the layout unless a human actually has to act. The login page already set `execution`, so only `appearance` was added there.

## Checks

`make lint` passes. `admin` typechecks clean. `web` typecheck reports the same 10 errors as clean `main` and none in the changed files — they are a pre-existing `EntryDelay.tsx` / `entryDelay.ts` case collision that only bites on a case-insensitive filesystem, so Linux CI is unaffected.

## Summary by CodeRabbit

- **New Features**
  - Added support for managed subscriptions, including clear status information and continued access to granted plans.
  - Subscription expiry is handled automatically.
  - Added an optional Beta notice displayed through a dashboard status pill and informational dialog.

- **Bug Fixes**
  - Corrected feature access checks for managed plans that do not use standard payment processing.
  - Improved CAPTCHA behavior by displaying challenges only when interaction is required and executing them when needed.

- **Documentation**
  - Documented the configuration option for the Beta notice.


meszmate:
Both addressed in 1e376da, and the Turnstile one was worse than described — thank you for catching it.

**Turnstile execution path.** You're right that `ref` never bound, and the reason is stronger than the typings: `Turnstile` is a plain function component, not `forwardRef`, and its `userRef` prop is typed `MutableRefObject<HTMLDivElement>` — the container div, not the widget. So neither `ref` nor `userRef` could ever reach `execute()` or `reset()`. The instance only arrives through `onLoad(widgetId, bound)`, which is what the login page already does correctly.

That mattered because of my own change: with the invalid `size="invisible"` the widget never rendered, and switching to `execution="execute"` would have traded that for a widget that renders and never runs. Both modals now bind in `onLoad` and execute from there.

The admin modal was additionally binding through `onAfterInteractive`, which only fires once a human has acted — with `appearance="interaction-only"` that may never happen, so a silent pass left the ref null. `onLoad` replaces it as the reliable hook; the interactive callback stays for the case where a challenge is shown.

**Missing effective plan.** Split, rather than treated as one failure. `perr != nil` is the instance being broken and now returns `errx.Internal` with the error captured. `plan == nil` is one workspace pointing at a plan row that is gone: reported through `errs.CaptureMessageContext` with the org and plan id, and the response still goes out without a plan. Failing the whole request there would take the billing page down for a data problem the user cannot act on, and the client already tolerates an absent plan — it falls back to the org row. Same distinction you had me draw between a repository failure and an unknown plan in #467.

Also in this push, unrelated to the review: a public-beta notice for preview deployments, shown once in a dialog and then as a header pill that reopens it, keyed off `WARMBLY_BETA_NOTICE` rather than a hostname so one built image stays reusable.

`make lint`, `admin` typecheck and `pnpm lint` clean; `web` typecheck matches clean main.

meszmate:
Both fixed.

**Mobile reachability** — a real hole, and the worst possible one for this feature: the pill was the only route back to the dialog, and a phone is exactly where someone dismisses it fastest without reading. It now sits outside the `hidden sm:flex` group, so it is present at every width.

**Docs lifecycle** — wrong on both counts, as you say. `config.js` is written when the container starts (or when the Pages build runs), so the value is baked, not read per load; the column now says a restart is needed. And "needs nothing unset" was backwards — production needs nothing *set*.

PR #483, opened 2026-09-13
feat: tester accounts, creatable from the admin panel

meszmate:
`AUTH_LOGIN_CODE` is instance-wide and read at boot. When one person outside the team has to sign in and cannot read this instance's mail — a vendor's reviewer during an OAuth verification, an auditor, a support engineer — the only lever was turning codes off for everyone, for as long as the review takes. Google's OAuth verification runs for weeks.

This narrows that to one named account, and puts the whole thing in the admin panel so letting a reviewer in is a form rather than a shell.

## Admin panel

A **Testers** page under Accounts:

- **Create** takes an address, an optional workspace name and a reason. It makes the account, its workspace and the exemption in one call, and shows the generated password **once** with copy buttons. Nothing stores it in a recoverable form; losing it means making another tester, which is cheap.
- **The list** is every account currently holding an exemption, with its reason and the date. This is the real point: the way this goes wrong is not creating one, it is forgetting it. An exemption taken out for a two-week review is still there a year later.
- **Revoke** drops the exemption and keeps the account, so anything it did stays attributable; it simply stops bypassing the code.

`warmblyctl user login-code-exempt` does the same from the CLI, and `warmblyctl status` lists exempt accounts on every run at warning severity.

## What the exemption does and does not remove

It removes the emailed code and nothing else. The password, the captcha and the sign-in risk assessment all still apply.

Three details that are deliberate:

- **Checked before the policy, not inside it.** `always` is precisely the mode that makes an outside review impossible, so an exemption that only worked under `new_device` would not solve the problem it exists for.
- **A read failure is not an exemption.** If the lookup errors the code is still demanded and the error is reported. Failing open would turn one bad query into a silent, instance-wide removal of the check.
- **An undeliverable transport still short-circuits first.** That guard exists so a deployment without working mail can sign in at all, and it stays ahead of this.

## Kept answerable

The reason, who granted it and when sit on the row, like `organizations.risk_override_*` and `subscriptions.managed_*`. A `CHECK` refuses an exemption with a blank reason; so do the CLI and the endpoint. Creating and revoking both go through `LogAdminAction`.

Creating a tester needs `ban_users` rather than `view_users`: both decide who can reach this deployment. Viewing the list needs only `view_users`.

## Tests

`login_code_exempt_test.go` covers the exemption holding under both `new_device` and `always`, a normal account still being challenged, a lookup failure still demanding the code, and the undeliverable-transport guard still winning.

`make lint` (150 migrations, no duplicates), `go test ./internal/app/auth/ ./internal/api/handler/`, and `pnpm typecheck` + `pnpm lint` in `admin/` all pass.

## Summary by CodeRabbit

- **New Features**
  - Added `warmblyctl user login-code-exempt` to exempt individual accounts from emailed login codes.
  - Exemptions require a reason, record the granting operator and date, and can be removed with `--clear`.
  - Exempt accounts bypass login codes even when the instance policy requires them.
  - Added an admin **Testers** page to create, view, and revoke tester accounts, with one-time credential display.
  - Status checks now report active login-code exemptions.

- **Documentation**
  - Added guidance for managing individual login-code exemptions.


meszmate:
All five addressed in 67af00c5.

**Permission (the one that mattered most).** You're right, and my justification in the PR body was hand-waving: "both decide who can reach this deployment" is not the same as authorising account creation. `ban_users` suspends an existing account; this mints one and hands back its password. New bit 22, `manage_testers`, in the catalog and mirrored in the admin TS enum, with `AllAdminPermissions` bumped to `1 << 23` and the compile-time live/retired assertion still holding. Viewing the list stays on `view_users`.

**Atomicity.** `CreateExemptUser` now writes the account and its exemption in one transaction. The orphan you described was the bad shape precisely because it was *invisible*: no exemption meant no tester-list entry, the address was taken so a retry failed, and whoever had the password could still use it.

I did not use a compensating delete. `HardDeleteUser` exists only inside the danger-zone service's scheduled-deletion flow, and exposing it as a public repository method to undo a form submission felt like the more dangerous change. Workspace creation stays outside the transaction because it runs through a service, but that failure is now benign and reported honestly: the account is already in the Testers list and the error says so, so nothing is hidden.

**CLI grantor.** Correct — `warmblyctl` connects straight to Postgres with no operator identity, so it was storing `NULL` and quietly defeating the column's purpose. `--by` is now required when granting, resolved to a user id. The admin panel already passed the acting admin, so it asks for nothing.

I took your advice on what *not* to do: no NOT NULL constraint on `login_code_exempt_by`. `ON DELETE SET NULL` matches how attribution references behave elsewhere here, and a constraint would make deleting a former operator fail while their exemption is still live.

**`rows.Err()`.** Fixed. Reporting the rows read before a halted scan as the complete list would understate what is exempt, which is the wrong direction for this particular check.

**UI error state.** Fixed, and this was the worst of the minors: rendering "None. Nothing on this instance is skipping the login code" when the query failed tells an operator the opposite of what they need to know. It now renders a distinct error saying the list is not authoritative.

`make lint` (150 migrations), `go test` across auth, instancecheck, handler and models, and `pnpm typecheck` in `admin/` all pass.

PR #505, opened 2026-09-14
fix: put the mailbox signature and the opt-out footer inside the container an HTML email was laid out in instead of after it, by locating that container with a new offset-keeping outline scan in internal/pkg/mailhtml and splicing into it, and centring the line on the card's own width when a builder export has no single container to sit in, so neither renders hard left in the page background any more (issue #462)

### LINKED ISSUE #462: Bug: Emails Sent in Full-Width Layout Instead of Centered Container
<br>
<br>
Below is an example of an email I sent using Warmbly. Currently, the email is being sent as a full-width layout.
Image

<br>
<br>
This is how the email should look with a properly centered email container, appropriate width, and spacing around the content.

Image

@meszmate

meszmate:
Closes #462.

## The bug

An email written in HTML is a card centred inside a full-width page table. `mailhtml.InsertBeforeBodyEnd` splices trailing content in just before `</body>`, which for that shape is **outside** the card: the mailbox signature and the opt-out line render flush against the left edge of the window, in the page background, with none of the styling the rest of the message carries.

That is exactly the screenshot on the issue, and it is two bugs in one place, since both `AddSignature` and `appendOptOut` append that way.

## The fix

A new `mailhtml.AppendToContent`, used by both, which places a fragment at the end of the message's **visible content**:

- It builds an element outline of the body from a source scan that keeps byte offsets, so the result is always the author's own bytes with the fragment spliced into one offset. It deliberately does not use `html.Parse`: that can only hand the tree back by re-rendering, and re-rendering a hand-written email rewrites markup nobody asked us to touch. A fuzz invariant asserts the splice property on every input.
- It descends from `<body>` while exactly one layout element holds the whole message, and appends inside that. The preheader and the 1x1 open-tracking pixel are skipped: both are children of `<body>` in a designed email (the pixel is added earlier in the send path), and counting either as layout stops the descent at the body and leaves the footer where it is today.
- A table with rows of its own gets a real `<tr><td colspan=N>` rather than a `<p>`, which every HTML parser hoists back out of a table and lands exactly where the bug put it. That cell picks up the side padding of the rows above it.
- A builder export (Unlayer, Beefree, Stripo) is a stack of full-width rows each centring a card of its own, so there is no single container to sit in. For those the fragment is centred on the card's own width, as a table rather than `max-width` + auto margins, which Outlook's Word engine ignores.
- Anything it cannot place with certainty falls back to `InsertBeforeBodyEnd`, so a campaign written in the visual composer is byte for byte what it was.

The preview and the test send share `finishBody` with the send path, so the composer shows the same placement a recipient gets.

## Verification

- `internal/pkg/mailhtml/append_test.go`: the centred card, preheader and pixel siblings, the appended table row, the builder export, conditional comments, omitted end tags, an unquoted attribute value ending in a slash, the four-value padding shorthand, plain bodies unchanged, and a splice-only invariant.
- `internal/tasks/optout_test.go`: the issue's own shape end to end, in send order (pixel, then signature, then footer), asserting both land inside the card's cell and in that order.
- Every new assertion was mutation-tested: the defect it guards was reintroduced and the check was watched to fail.
- The splice invariant is in the existing fuzzer; ~4M executions across runs, no failures.
- `make fmt`, `make lint`, `gofmt -l` clean. `pnpm types:check` and `pnpm lint` clean in `docs/`.

## Docs

`guides/unsubscribe.mdx` gains a "Where it appears" section, and the signature paragraph in `guides/mailboxes.mdx` points at it.

## Not in this PR

The issue's second comment asks to replace the composer with GrapesJS. That is a separate product call, not a fix for this bug, and worth answering on the issue rather than folding in here.

## Summary by CodeRabbit

* **Enhancements**
  * HTML email signatures and opt-out links now appear within the message’s content layout, preserving alignment, styling, and spacing across centered, full-width, table-based, and visual-composer emails.
  * Plain-text email bodies remain unchanged.
* **Documentation**
  * Added guidance explaining where signatures and opt-out lines appear in different email layouts.
* **Bug Fixes**
  * Improved placement around tracking pixels, conditional content, tables, and responsive email structures.


meszmate:
@coderabbitai review

meszmate:
Pushed two more fixes found while re-reviewing, both mutation-tested:

- The outline read `<td><p>One<td>Two` as the second cell nested inside the first, because `closeNamed` only walked through inline elements the author had left open and not through the ones HTML makes the end tag optional for. A table written that way reported one column, so the appended row did not span it.
- A trailing slash inside an *unquoted* attribute value was read as a self-closing tag, so `<td background=/bg.gif/>` never opened and everything written inside it became a sibling, moving the end of the content.
- The four-value `padding` shorthand took its left from the second value, so the appended row sat out of line with the cell above it.
- `isCentred` matched `margin:0 auto` as text, which MJML does not write: it emits `margin:0px auto`. A multi-section MJML export therefore got no centring wrapper and still went out hard left. The shorthand is read now rather than matched.

Also checked the placement by hand against Mailchimp, Stripo, Unlayer, MJML (one section and several) and Foundation for Emails; MJML now has a regression test of its own.

meszmate:
@coderabbitai review

meszmate:
Force-pushed to `6e0991fb`. The previous head had swept in eight files of unrelated in-progress warmup work (`internal/tasks/email_task.go`, the partner-selection tests, `AGENTS.md`, `docs/public/openapi.json` and two other docs pages) from a shared worktree. Nothing was lost; those changes are back where they belong and this branch is the ten files it should be. Re-verified the clean commit in an isolated checkout: `gofmt`, `make lint` and the `mailhtml` + `tasks` suites all clean.

meszmate:
@coderabbitai review

meszmate:
@coderabbitai review

meszmate:
Two more from a final self-review, both mutation-tested:

- `contentWidth` could size the line on a call-to-action button. In a card whose width lives in a stylesheet nothing on the insertion point's chain declares one, so a centred `width="300"` button was the only candidate in sight and the opt-out went out a third the width of the copy above it, centred. It now stops at the outermost centred box rather than counting modules nested inside it, and the floor rose from 200px to 400px, which is above button territory and below the narrowest real template.
- `sidePadding` copied the author's own padding value into a `style` attribute of our own without checking it was a length. A garbled `style='padding:0 "x'` produced `style="padding-left:"x;..."`, which closes itself early and breaks the cell. Only a real length is copied now; anything else is dropped and the row ships unpadded.

Also documented `closeTo`'s invariant (n has to be `cur` or one of its ancestors) since nothing said so and neither caller can violate it today.

Head is `609eab64`.

meszmate:
@coderabbitai review

meszmate:
Thanks — all three were real. Fixed in `de74717a`, each mutation-tested (defect reintroduced, check watched to fail):

**`noscript`** — agreed, and the reasoning is stronger in mail than in a browser: clients run no scripts, so the fallback content is always what the reader sees. Removed from `nonLayoutElements`; a body with one now falls back to appending at the end, which keeps the opt-out last.

**`colspan`** — confirmed. A table whose every row uses `colspan` (a 3-wide banner over a `colspan=2` + `td` row) reported two columns, so the appended row stopped short of the copy. Counting grid columns now, via a `cellSpan` helper that falls back to 1 for anything unreadable and ignores a span past HTML's own cap of 1000.

**The quadratic ancestor walk** — measured before fixing: 250 KB of `<span>×k` + `<p></p>×k` took **4.03s**, and 2 MB took minutes. Worth noting `openAncestor` had the same shape, so a stray-end-tag body was quadratic too and a bound on `closeNamed` alone would have left that. Fixed at the source instead: `maxOutlineDepth` (64) stops the outline pushing past a nesting depth no real template reaches, which bounds every ancestor walk in the file at once. Same input is now **0.03s**, and 2 MB is 0.65s. There's a test asserting it stays under 2s, which takes 6s without the fix.

Also verified the result on top of `d74d5e68`: builds, `gofmt` clean, both suites and golangci-lint green.

meszmate:
@coderabbitai review

meszmate:
CI caught my own test, not the fix: `TestOutlineStaysLinearOnUnclosedNesting` asserted a wall-clock threshold, and CI runs `go test -race -covermode=atomic`. The same input that takes 0.03s here takes 2.8s locally under those flags and 5.6s on the runner, while being just as bounded — the assertion was measuring the machine.

Replaced with a deterministic one on the invariant that actually makes the walk cheap: every ancestor walk runs from an element to the root, so the test now asserts the outline's deepest node stays within `maxOutlineDepth`. Instant, environment-independent, and still mutation-sensitive — without the bound it reports "nesting reached 20001, past the 64 bound".

Verified under CI's exact flags (`-race -covermode=atomic -cover`): `mailhtml` and `tasks` both green. Head is `9f255312`.

PR #506, opened 2026-09-14
Erase everything a disconnected mailbox leaves behind, and make it deletable from somewhere obvious

meszmate:
## What this fixes

Disconnecting a mailbox deleted rows and stopped there. Three things survived it:

1. **The OAuth grant stayed live at the provider.** Deleting our copy of a refresh token is not revocation — Warmbly kept appearing in the customer's Google third-party access list forever. Google's API Services User Data Policy requires that a user be able to have their data deleted.
2. **Every message body stayed in object storage.** The `unibox_emails` rows cascaded; the `.emsg` files they point at did not.
3. **Nine tables named a mailbox with no foreign key**, so their rows outlived it: `warmup_received`, `warmup_tampering_events`, `email_message_map`, `email_history_ids`, `email_delta_links`, `warmup_pending_engagements`, `ai_thread_drafts`, and `decision_log`. Two of those hold the subjects and message ids of mail that arrived in somebody's inbox.

There was also no obvious way to delete one mailbox: it was reachable only by ticking a row's checkbox and finding the floating selection bar.

## How erasure works

The delete removes the rows in one transaction and writes down what it cannot do there — an HTTP call and an object-store walk, neither of which may be lost if the process handling the request goes away. `mailbox_erasures` is that record, and a job (`mailbox_erasure`, every minute) works it off:

- **Revoke.** Gmail refresh tokens go to `https://oauth2.googleapis.com/revoke`, which invalidates the token and every access token derived from it. A `400 invalid_token` means already-revoked and counts as done. Microsoft publishes no per-application revocation endpoint (the only Graph call that ends sessions signs the person out of every app), so Outlook tokens are destroyed locally and the docs tell the customer where to remove the app.
- **Erase.** `Store.DeletePrefix` removes everything under `users/<user>/emails/<mailbox>/` in one call rather than one per message. S3 pages through `ListObjectsV2` + `DeleteObjects`; the filesystem store does one `RemoveAll`. `CheckPrefix` refuses anything shallower than four segments or without a trailing slash, because `DeletePrefix("")` empties the bucket and `…/emails/abc` (no slash) also matches `…/emails/abc-other/`.

Finishing **deletes the queue row** rather than stamping it done: the row holds the address the customer asked to have forgotten, so keeping it as a receipt would be keeping the thing being erased. The audit log already records the deletion. Failures retry with a widening backoff capped at 6 hours and are never abandoned; anything outstanding past 24h is reported as the job's error, so it surfaces in the admin panel's job list.

Rows move only as the mailbox's own transaction allows, so a delete that fails (a `503` when the worker can't be told) queues nothing — erasing a live mailbox's grant would be far worse than not erasing.

## Workspace deletion never worked

`HardDeleteOrganization` is one `DELETE FROM organizations`, documented as "relying on existing ON DELETE CASCADE FKs". Four of them had **no delete action at all** (`campaigns`, `contacts`, `sequences`, `email_accounts`), so it raised a foreign key violation for any workspace that had ever held one of those — every workspace anyone would ask to delete. A scheduled deletion ran its 30-day grace period, sent its warning emails, and then failed every tick forever with the workspace still there. All four now cascade. `sessions.current_organization_id` becomes `SET NULL`, because a session belongs to the person, not the workspace they happened to be viewing.

## Known and deliberately not fixed here

**Account closure still cannot complete**, for an unrelated reason: about twenty foreign keys into `users` are `NO ACTION`, including `scheduled_deletions.requested_by_user_id`, which the deletion request itself writes. Fixing it needs a separate decision per constraint about what survives a person leaving — an admin audit trail should keep its rows and lose the author, a teammate's workspace must not be cascaded away because its owner left. That is its own piece of work. This PR makes sure the failure stays clean: the enqueue shares the delete's transaction, so nothing is erased while the mailbox is still connected, and there is a test holding that line.

## UI

- Each row's **More** menu now drops **Mailbox settings** and a red **Disconnect mailbox**. It previously jumped straight to the Settings tab; `AGENTS.md` is updated to describe the new convention.
- The mailbox drawer's **Settings** tab gained a **Danger zone**, matching the workspace danger zone in settings.
- All three confirmations now say what is actually lost and that it cannot be undone, and point at switching the mailbox off instead.
- `useRemoveEmail` invalidated only `["emails", id]`, so a drawer delete would have left the mailbox allowance counter still claiming the workspace was at its limit. It now invalidates the `["emails"]` prefix and account health, matching the bulk path and the realtime spine.

## Docs

`guides/mailboxes.mdx` (what disconnecting erases, and where the control is), `development/data-control.mdx` (the queue, and which store holds what), `api/openapi.json`, the `warmbly mailbox remove` help, and a new **Google and Microsoft mailbox data** section in the privacy policy carrying the Limited Use disclosure Google requires for the restricted Gmail scopes this app uses. `mailbox_erasures` is registered in `orgtransfer`'s `ExcludedTables`.

## Verification

`make lint` (including `check-migrations`), `gofmt -l`, `go vet`, web `pnpm typecheck` + `pnpm lint` (0 errors, same 84 pre-existing warnings), docs `pnpm types:check`, site `pnpm build`. Live tests ran against an isolated database with the new migration, plus a full down-and-up cycle of 000157.

Every new assertion was mutation-tested by reintroducing the bug it guards — including each of the new foreign keys individually. One warmup assertion passed under mutation because the fixture used the same mailbox on both sides of a receipt, so it was rewritten to test both directions. The `internal/tasks` failure on `TestLiveRepairedVerificationRestoresRouting` is pre-existing, confirmed by stashing this branch and reproducing it on main.

## Summary by CodeRabbit

- **New Features**
  - Added mailbox **More** menus with options for mailbox settings and disconnecting.
  - Added a disconnect action in mailbox Settings, with confirmation and success/error feedback.
  - Mailbox disconnection now removes imported message data and warmup history, and revokes provider access where supported.

- **Bug Fixes**
  - Mailbox deletion now reliably cleans up related data and removes orphaned thread labels and snoozes.
  - Cleanup continues asynchronously with retries if provider or storage services are temporarily unavailable.

- **Documentation**
  - Updated mailbox deletion guidance, API documentation, development documentation, and privacy disclosures.


meszmate:
Pushed a revision after a self-review pass, before the review lands:

- **The migration locked several live tables at once.** Every foreign key is now added `NOT VALID` (binds new writes and makes the delete actions live immediately, no scan under the lock), with the scan moved to `000158` in its own transaction under `SHARE UPDATE EXCLUSIVE`. golang-migrate sends a whole file as one implicit transaction, so validating in `000157` would have sat behind its own `ACCESS EXCLUSIVE` locks on `contacts` and `campaigns`. The `unibox_emails` index moved to `000159` as `CREATE INDEX CONCURRENTLY`, alone in its file, matching `000137`'s precedent.
- **Orphan cleanup uses `NOT EXISTS` rather than `NOT IN`** — plans as an anti-join, and `NOT IN` silently matches nothing if the subquery ever yields a NULL.
- **The job wrote two rows it immediately deleted.** The half-done stamps now only persist when there will be a retry, which is the only thing they are for.
- **A batch could outrun its own lease** (50 mailboxes x a slow provider > 10 minutes), letting another backend claim rows still being worked. There is now a batch budget of half the lease.
- **A cancelled context counted as a failed erasure**, burning retry attempts and widening the backoff on every redeploy. Shutdown now stops the loop instead.
- **An unreadable backlog count returned nil**, showing the job green while nothing knew if anything was stuck.
- `ON CONFLICT` now resets the progress stamps too, so a re-enqueued mailbox can never skip revoking a new grant because an older erasure was recorded as done.

All three migrations verified with a full down-and-up cycle, and the two new guards were mutation-tested.

meszmate:
Added an end-to-end live test for the erasure path (`internal/jobs/mailbox_erasure_live_test.go`): a real mailbox row with a real OAuth token, three real message bodies in a real filesystem blob store, the real `Delete`, then the real job.

The unit tests stub the repository and the store, so they can show the job's decisions but not that the pieces fit. This asserts the delete writes a row the job can claim, that the prefix it recorded is the one the bodies were actually written under, that a neighbouring mailbox's mail survives, and that the queue row is gone afterwards. A second test asserts a following pass finds nothing left to do, so a job that failed to clear its row cannot keep revoking and re-walking forever.

Mutation-tested by pointing `StorageEndpointEmailBody` at a different prefix than `StorageEndpointMailboxPrefix`: the test fails with the three orphaned bodies named. That drift is the silent failure this whole change exists to prevent, and nothing else in the suite could see it.

meszmate:
Worked through all nine review findings. Two of them (the migration's locking, and the swallowed `PendingOlderThan` error) were already fixed in `e08f9a4` and CodeRabbit marked them addressed. The other seven are fixed here, plus a rebase onto current main.

**The one where the review was right and I was wrong.** I had `revoke` complete the erasure when a token could not be opened and the provider answered `400 invalid_token`, reasoning that the stored value is all we will ever have so retrying cannot improve it. That is wrong: a `400` is the same answer Google gives to ciphertext sent under the wrong key, which is exactly what a restore with a mismatched `CREDENTIALS_ENCRYPTION_KEY` produces. Completing there records a grant as revoked while it is still live on the customer's account, which is the one outcome this job exists to prevent. It now fails and stays outstanding, naming the likely misconfiguration. My concern about retrying forever is already handled: the backoff caps at six hours and `PendingOlderThan` surfaces the row as the job's error after a day. Visible and unfinished beats finished and wrong. A `200` on an unopened token still completes, because acceptance is proof regardless of what we sent.

**`BLOB_PROVIDER=brokered` on the backend.** It cannot enumerate, so every erasure would have sat in the queue with the customer's mail still in the bucket. The backend now refuses it at boot with a message saying why. The probe uses the empty prefix, which every real backend rejects with `ErrUnsafePrefix` before touching anything, so only the brokered store answers `ErrUnsupported`.

**The rollback destroyed pending erasures.** `DROP TABLE mailbox_erasures` in `000157`'s down discarded the only remaining copy of a deleted mailbox's refresh token and blob prefix, making revocation and deletion permanently impossible. It now raises an exception naming the count and telling you to let the queue drain. Verified both ways: refuses with a row present, succeeds once drained.

**`last_error` truncation split multibyte characters.** A provider error body is arbitrary remote text, so cutting at 500 bytes could produce invalid UTF-8, which Postgres rejects, losing the attempt count and backoff with it. Now truncates by rune after dropping invalid input, with tests over multibyte and invalid-byte inputs.

**Outlook revocation was over-promised in three places.** Microsoft publishes no way for us to remove a single application, so the row confirmation, the bulk confirmation and the CLI success message all claimed something we cannot do. Each is now provider-specific, and the bulk one builds its sentence from the providers actually selected so a mixed selection never rounds up to the stronger claim.

**The CLI pointed at a command that would fail.** `warmbly mailbox edit MAILBOX_ID` returns the body-required error; the help now shows `--input '{"status":"inactive"}'`.

**The privacy policy promised deletion on account closure.** Corrected to name disconnecting a mailbox and deleting a workspace, both of which do work.

One thing I did **not** change: two pre-existing retention bullets still say data is deleted after account closure. Those predate this PR and the right fix is making account closure work, not weakening the published policy, so I have left them for that separate piece of work rather than editing legal copy on the way past.

meszmate:
Also pinned the property the new boot guard depends on. `cmd/backend` probes its blob store with the empty prefix to learn whether it can erase a mailbox at all, which only works because a real backend answers `ErrUnsafePrefix` (refusing before touching anything) while the brokered one answers `ErrUnsupported`. If `BrokeredStore.DeletePrefix` ever validated the prefix first, both would answer `ErrUnsafePrefix` and the guard would silently stop catching the misconfiguration it exists for. There is now a test for exactly that, mutation-verified by adding the check and watching it fail.

@coderabbitai review

meszmate:
CI's Migrations job failed on the previous head, correctly: main merged #508 carrying `000157_drop_warmup_spam_score` while this branch also had a `000157`. Rebased and renumbered mine to `000158`/`000159`/`000160`, since main's is already released and renumbering that one would make deployments re-apply it. Cross-references in the migration comments and the live tests were updated with them.

That collision surfaced an interaction worth checking rather than assuming. #508 rewrote the `warmup_reputation_mirror` trigger, and this PR makes `email_accounts` cascade from `organizations` for the first time, so deleting a workspace now unwinds organizations -> email_accounts -> warmup_pool_participants alongside `warmup_reputation_ledger`, which hangs off the organization by its own key. The existing workspace tests use a bare mailbox and would not have noticed if one of those links stopped resolving, and a foreign key that does not aborts the entire deletion.

There is now a test for it, with a pooled mailbox carrying a blocked standing and a mirrored ledger row. I also checked whether the trigger itself runs during the cascade: it fires on insert and update only, so it does not, and I corrected the test's comment rather than leave it claiming coverage it does not have.

Full Go suite, `make lint`, and the live suite all green against a database rebuilt from migrations including main's #508.

PR #509, opened 2026-09-14
Rebuild the unified inbox: three columns, a flattened rail, an on-the-spot filter popover, and real loading states

meszmate:
Rebuilds the unified inbox. The old screen had a metric strip nobody clicked, a rail with two competing row styles, list rows built around an avatar, a thread header that led with the word "Thread", and a full-height filter sheet with an Apply button. This replaces all of it and adds the loading behaviour the screen never had.

## Layout

The top metric strip is gone. Its six counters duplicated numbers already in the rail, so the page is now three columns and nothing above them. `UniboxHeader.tsx` is deleted, along with the never-imported `UniboxFilters.tsx`.

**Scope rail.** One visual language for every row: a 15px icon, the label, a bare tabular count that turns blue only when it is an unread count. Two groups with no section headers (All mail, Inbox, Unread, Awaiting reply, Agent drafts, Snoozed, then Drafts, Sent, Scheduled, Archive, Spam, Trash), then Mailboxes, Labels and Tags. White instead of a second grey surface. The folder three-dot menu now appears in place of the count on hover rather than shifting the row.

**Conversation rows.** Three tight lines, no avatar, so the eye reads one column of names. Unread is a blue dot in the gutter plus weight plus a blue timestamp, in place of the left bar. Labels sit at the end of the subject line. The owning mailbox shows only when the workspace has more than one.

**List header.** A real title row (view name, count, filter button) over a rounded search field with a `/` hint. The keyboard cheat-sheet footer is gone; the `?` modal already lists every shortcut.

**Thread header.** The subject as a title over one meta line carrying the message count, mailbox, labels and who else is viewing. All actions are icon-only under tooltips. The "THREAD" eyebrow, the separate message-count bar and the "Hover any message to reply" hint are removed. Collapsed messages are one line of sender plus preview; bodies indent to the text column.

**Removed on purpose:** Today and This week are no longer rail rows. The routes still resolve, and the filter popover's date range covers the same slice.

## Filters

The slide-over sheet is replaced by a 300px popover under the filter button, applied on the spot, with no Apply, Cancel or Reset footer. Segmented controls for read state, date (Any / Today / 7d / 30d / Custom) and sort; a debounced sender field; tinted chips for labels; a mailbox checkbox list with tag chips above it.

Two things the sheet got wrong are fixed. It offered every control in every view, so on the Unread view you could pick "Read" and get a contradiction; the popover now hides whatever the current view already fixes, comparing against a new `baseParams` prop that carries what the scope alone would query. And an active filter was invisible once the sheet closed; each one now shows as a chip under the search box that removes itself with one click.

## Loading

- A 2px indeterminate bar along the top of the list during a scope, search or filter change, delayed 150ms so a fast response never shows it. Background refetches from realtime events do not trigger it.
- Stale rows stay on screen at half opacity until the new set lands, then swap in one go. The row container is keyed on the query whose data is showing, so a full swap remounts instead of animating fifty rows at once.
- Row-shaped skeletons with varied widths for first load, for the next page, and for the thread reader, which now shows the header and message layout the real thread lands in.
- Archive, delete and snooze remove the row and close the reader immediately through a new `listCache.ts` helper. The request runs behind the toast; a failure re-reads the list, which is the rollback. Undo still works.
- Rail counts tick rather than jump.

## Kept in sync

- `docs/content/docs/guides/unibox.mdx` describes the new rail, list, header and filters.
- `site/src/components/UniboxMock.astro` mirrors the new screen 1:1.

## Verification

| Check | Result |
| --- | --- |
| `pnpm typecheck` (web) | pass |
| `pnpm lint` (web) | 0 errors |
| `pnpm test:run` (web) | 288 passed |
| `pnpm build` (web) | pass |
| `pnpm build` (site) | pass |
| `pnpm types:check` + `pnpm lint` (docs) | pass |

Also exercised by hand against a seeded local stack: 41 conversations across all four time buckets, every folder, six labels, three tags, snoozes, agent drafts and scheduled sends.

## One observation, not addressed here

The overview's Awaiting reply count matches a bare mailbox address against `from_addr`, but synced mail carries a display name, so the count reads zero on seeded data. I could not check what real synced mail stores, so I have left it alone rather than guess.

## Summary by CodeRabbit

- **New Features**
  - Added popover-based filtering with removable filter chips for unread status, dates, senders, labels, mailboxes, tags, and sorting.
  - Added clearer scope navigation for views, folders, mailboxes, labels, and tags.
  - Added mobile scope switching and improved scheduled-message controls.
  - Conversations now support grouped rows, label chips, mailbox context, and clearer unread indicators.
  - Added optimistic removal of filed or snoozed conversations from active lists.

- **Style**
  - Refreshed conversation, thread, contact, loading, and empty-state layouts for a cleaner experience.


meszmate:
@coderabbitai review

meszmate:
@coderabbitai review

meszmate:
@coderabbitai review

Issue #470, opened 2026-09-12
Out-of-office and auto-replies: hold the contact until they are back, plus a manual per-contact pause instead of unsubscribe


Skylinerffm97:
Self-hosted on Railway. Verified the behaviour against main at 017f4cf today.

## What happens

A recipient's mailbox answers a step with an out-of-office auto-reply, often as a fresh email rather than a reply in the thread. warmbly classifies it correctly (replyclassify → out_of_office, matched to the contact by sender address) and, by design, does not treat it as a human reply: internal/app/consumer/event_new_email.go never sets replied_at for automated replies, so stop_on_reply does not fire. The result is that the follow-up goes out on schedule while the person is still away, and the sequence is over before they are back.

Two cases from my instance this week:
- 04.09. "Automatische Antwort: …", no return date in the body
- 07.09. "Automatic reply: …" with "I will reply to your email on my return on the 8th September"

There is no per-contact pause. The only levers are the "Subscribed" toggle (removes the person from every campaign, permanently) and the suppression list. Neither is the right tool for "this person is on holiday for a week".

## What I would expect

1. On out_of_office, park the contact's next step. If the auto-reply carries a return date ("back on", "return on", "zurück am", "ab dem", ISO and DD.MM.YYYY dates), resume at return date + 1 business day. If no date can be parsed, hold for a configurable default (e.g. 7 days) and resume. Held time should not count against the step's wait.
2. Show it on the lead: "out of office until …" in the campaign Leads tab and the contact drawer, with "resume now" and "stop" next to it.
3. The same hold available manually: pause this contact in this campaign until a date, without unsubscribing them or removing them from the campaign. Instantly and Smartlead both have this and it is the one thing I miss most day to day.

## Related: non-English defaults

reply_intent.out_of_office_keywords ships English-only ("out of office, ooo, vacation, automatic reply"), and the subject rules in internal/app/replyclassify/headers.go are English-only too, so a German auto-reply is only caught when it carries Auto-Submitted or vendor headers. Suggested defaults: "Abwesenheitsnotiz", "abwesend", "außer Haus", "nicht im Büro", "im Urlaub", "zurück am", "Automatische Antwort". Glad to supply lists for other languages.

If you would rather take this yourself I am happy to leave it with you; if you want a PR I would start with the hold/resume and the keyword defaults and keep the date parsing conservative.

PR #520, opened 2026-09-15
feat: hold an out-of-office contact in every campaign they are a lead of, not only the one that was answered

### LINKED ISSUE #518: An out-of-office hold covers one campaign, but being away is a property of the person
Raised by @Skylinerffm97 on #512.

An out-of-office hold is written against **one** campaign: the one the reply resolved to. But being away is a property of the person, not of one sequence. A contact who sits in two live campaigns gets one of them held and keeps being mailed by the other, which is issue #470's own failure narrowed to the second campaign.

## How the campaign is resolved today

`ProcessIncomingReply` (`internal/app/advanced/service.go:1037`) attributes an inbound message in two steps:

1. `In-Reply-To` pointing at a Message-ID we sent resolves the exact campaign, contact and step.
2. Failing that, the normalised `From` address is matched against the organisation's contacts, and the campaign is the contact's **most recently sent** one (`GetLatestCampaignSequenceForContact`, `internal/repository/pg_campaign_progress.go:1111`).

`holdForOutOfOffice` then calls `HoldLead(campaignID, contactID, ...)` for that single campaign.

Both paths are wrong for a multi-campaign contact, and path 2 is wrong in a second way: it orders by `sent_at` with no status filter, so the campaign it picks can be a **completed** one whose last step happens to be the most recent send. The away notice then holds nothing that was still going to send.

## What it should do

An out-of-office holds the contact in every campaign they are still a lead of. The existing guard semantics carry over unchanged: an automatic hold never overrules a person's own pause, and a live automatic hold is only ever extended, never cut short.

Out of scope, deliberately: an away notice arriving from an address we never mailed (a delegate, a shared mailbox, an alias) matches no contact and writes no hold. There is no reliable signal tying that message to a lead, and guessing would pause a contact who is not away, which is a worse failure than not holding. It stays in the unibox.

## Acceptance

- an out-of-office reply holds the contact in every non-completed campaign where they are a lead
- a member's own manual hold in any of those campaigns is left alone
- a longer live automatic hold is not cut short by a shorter new one
- resume stays per campaign, unchanged
- the campaigns guide says the hold covers every campaign

meszmate:
Closes #518. Raised by @Skylinerffm97 on #512.

An out-of-office hold was written against one campaign: the one the reply resolved to. Being away is a property of the person, not of one sequence, so a contact you are running two campaigns to had one of them held while the other kept writing to the same empty desk. That is #470's own failure, narrowed to the second campaign.

## The change

`HoldLeadEverywhere(contactID, until, reason, source)` holds the contact in every campaign they are still a lead of, in one statement, and returns the campaigns it actually held so the caller can tell a hold from a refusal. `holdForOutOfOffice` calls it instead of `HoldLead` and no longer needs the campaign id at all.

One statement rather than a read-then-write loop, so two away messages landing at once cannot each decide against a snapshot the other has already moved: the guard is evaluated against the row it updates, exactly as in `HoldLead`. That guard is now `automaticHoldGuard`, shared by both writes rather than spelled out twice, so a member's own pause and a longer running automatic hold survive per campaign and not just in the single-campaign path.

Completed campaigns are left out, since nothing in them will send again. Draft and paused ones are held: the contact is still their lead, and an away message that lands the day before a campaign starts is about the same desk. Every automatic hold carries an end, so one written into a campaign that never launches expires on its own.

No migration and no new index: `idx_campaign_leads_contact` already covers the lookup. The manual pause, resume, and the `hold` object on every read stay per campaign and are untouched, which is the right shape for them: pausing by hand is a decision about one sequence, not a report that the person is away.

## Deliberately not fixed

The second hole in #512's thread: an away notice arriving from an address we never mailed (a delegate, a shared mailbox, an alias) matches no contact and writes no hold. There is no reliable signal tying that message back to a lead, and guessing would pause a contact who is not away, which is a worse failure than not holding. It still lands in the unibox. Said as much on the issue rather than leaving it to be rediscovered.

Also left alone: `GetLatestCampaignSequenceForContact` orders by `sent_at` with no status filter, so an untethered reply can be attributed to a completed campaign. It decides which campaign the reply classification is stored against, and changing it would move the reply-branch path for a gain this PR already delivers another way, since the hold no longer depends on which campaign was picked.

## Verification

`make lint` (with `check-migrations`), `pnpm typecheck` + `pnpm lint` in `web/`, `pnpm types:check` + `pnpm lint` in `docs/`, and the full Go suite on a freshly migrated scratch database.

Four new live tests, each mutation-checked against the fix reverted:

- holding only the resolved campaign again fails `…CoversEveryCampaignTheContactIsIn`
- dropping the guard fails `…LeavesAPersonsOwnPauseAlone` and `…NeverShortensALiveAutomaticHold`
- including completed campaigns fails `…SkipsACompletedCampaign`
- renaming the recorder's `HoldLeadEverywhere` so the service falls through to the embedded interface panics `TestHoldForOutOfOffice`, which is what pins the service to the person-wide write rather than the single-campaign one

`internal/tasks`' `TestLiveRepairedVerificationRestoresRouting` fails with the same `read completion log: no rows in result set` it fails with on `main`; it is not from this branch.

## Docs

The campaigns guide now says the hold covers every campaign and why the manual pause does not, the sequences guide's auto-reply note says the same, both agent skills say it so neither tells an agent to pause a lead that is already held, and the **Settings > Sending > Out of office** description says how far the hold reaches.

## Summary by CodeRabbit

- **New Features**
  - Out-of-office replies now place a lead on hold across every active campaign they belong to.
  - Completed campaigns remain unaffected.
  - Existing manual pauses and longer automatic holds are preserved.

- **Documentation**
  - Updated campaign, sequence, API, CLI, and sending-safety guidance to clarify out-of-office hold behavior.


meszmate:
@coderabbitai full review

meszmate:
@coderabbitai full review

Issue #410, opened 2026-09-10
Feature request: native ASN-based filtering of Microsoft 365 security-scanner false opens and clicks


akminmkannan:
### Title

Feature request: native ASN-based filtering of Microsoft 365 security-scanner false opens and clicks

### Body

**Problem**

Microsoft 365's security infrastructure (Safe Links / Safe Attachments / ATP-style scanners) proactively fetches tracking pixels and tracked links in inbound mail, independent of whether the recipient actually opens or clicks. This inflates open/click analytics with false positives and undermines warmup/deliverability metrics that rely on genuine engagement signals.

**Evidence**

We captured a clean example: the same Microsoft-owned IP generated both an open request (`/t/o/...`) and a click request (`/c/...`) for the same tracked email, using two different human-looking User-Agent strings, while the actual recipient confirmed they had neither opened nor clicked. The IP resolved to Microsoft's public ASN `8075`.

**Workaround implemented (infra layer, not in Warmbly)**

We set Cloudflare in front of our tracking hostname to inject the requester's ASN as a header:

```
MK-CF-ASN = ip.src.asnum
```

and added an Apache rewrite rule to suppress open-pixel requests specifically from known Microsoft-owned ASNs:

```apache
RewriteEngine On

RewriteCond %{HTTP_HOST} ^<tracking-hostname>$ [NC]
RewriteCond %{HTTP:MK-CF-ASN} ^(8074|8075|12076)$
RewriteCond %{REQUEST_URI} ^/t/o/ [NC]
RewriteRule ^ - [R=204,L]
```

Validated behavior:

```
ASN 8074  -> 204 (suppressed, request never reaches Warmbly)
ASN 8075  -> 204 (suppressed)
ASN 12076 -> 204 (suppressed)
ASN 8560  -> 200 (normal, reaches Warmbly as usual)
```

This reliably filters false opens originating from Microsoft's scanner infrastructure before they ever reach Warmbly's tracking endpoint.

**What's still unsolved: click tracking**

We deliberately did *not* apply the same suppression to `/c/...` (click) requests, because a real human click can legitimately arrive via Microsoft's infrastructure (e.g. Safe Links rewrites the URL and proxies the click through Microsoft's own servers even for a genuine user action). Returning `204` on `/c/` would break the redirect for real users, so blunt ASN suppression isn't safe there the way it is for the open pixel.

**Feature request**

[...]

PR #526, opened 2026-09-15
feat: resolve scanner ASNs from a GeoLite2-ASN database and ship the mail-security vendor ASNs enabled as probable (#440)

### LINKED ISSUE #440: Resolve scanner ASNs from a GeoLite2-ASN database so asn: catalogue entries work without a Cloudflare transform rule
## The gap

`asn:` entries in `tracking/scanner-networks.txt` do nothing unless `TRACKING_SCANNER_ASN_HEADER` names a header a trusted proxy writes with the source ASN. In practice that means Cloudflare with a transform rule on `ip.src.asnum`. An instance behind nginx, Caddy, a cloud load balancer, or exposed directly gets no ASN matching at all, and the catalogue silently warns about it at boot:

> Scanner catalogue has ASN entries but TRACKING_SCANNER_ASN_HEADER is unset, so none of them can match

This matters more after #438, because the ASN half of the catalogue is where the mail-security vendors live: Proofpoint (`22843`, `26211`, `52129`), Mimecast (`30031`, `39588`, `42427`, `60492`) and Cisco Secure Email (`16417`). Those vendors publish no short list of scanner CIDRs the way Microsoft and Barracuda do, so the ASN is the only practical handle on them, and today it is reachable only by operators who happen to run Cloudflare.

## Proposal

Resolve the source ASN from a MaxMind GeoLite2-ASN database, so `asn:` entries match with no edge configuration. The header stays supported and keeps precedence where it is set.

GeoLite2-ASN is a free download on the same licence as the GeoLite2-City database this repo already uses, so it adds no new vendor relationship, only a second file.

## The open decision: where the lookup goes

**A. In the tracking service (recommended).** It already owns the catalogue (`include_str!` of `scanner-networks.txt`) and already does the classification in `ScannerNetworks::classify`. An ASN database there makes `asn:` entries work with no event-schema change, no consumer change, and no second copy of the catalogue. Cost is a new Rust dependency (`maxminddb`) and a mounted file in the tracking image.

**B. In the consumer.** It already receives the raw client IP (`event.ClientIP`, used by `originOf`) and already holds a MaxMind reader. But the catalogue lives in the Rust binary, so the `asn:` entries would have to be parsed again in Go, leaving two parsers over one file. It also splits scanner classification across two services, when `isScannerSource` exists precisely because the edge is supposed to own that verdict.

A is cleaner. The one argument for B is the existing split noted in `docs/content/docs/development/events.mdx`, that the backend is where a user agent and IP become device and location. That reasoning is about not putting the IP in a durable stream on the way to enrichment; it does not apply here, because the tracking service is the edge and already has the address in hand.

## Verified constraints

- `geoip2-golang/v2` exposes `Reader.ASN(netip.Addr)`, so option B is a few lines. But `ASN()` returns `InvalidMethodError` when the open database is not an ASN database, and `GEODB_PATH` points at GeoLite2-**City**. Either option needs a **second** database file and reader; the existing one cannot be reused.
- `tracking/Cargo.toml` has no maxminddb crate today (`ipnet` is there for the CIDR matching).
- The lookup is an mmap'd radix-tree read, microseconds, and the file is mmap'd rather than resident, so the tracking service stays light. It is on the pixel hot path, so it should still be behind the same "absent file is tolerated" rule as `GEODB_PATH`: a missing or unreadable database disables ASN matching with a logged note and never fails a request.

## Acceptance criteria

- A new variable (suggest `TRACKING_SCANNER_ASN_DB`) points at a GeoLite2-ASN database; unset or unreadable disables native ASN matching and logs once at boot, exactly as `GEODB_PATH` does today.
- With it set and no `TRACKING_SCANNER_ASN_HEADER`, an `asn:` entry matches.
- Where both are configured, the trusted header wins, so an operator who already writes it keeps the behaviour they have.
- The header is still read only from a peer in `TRACKING_TRUSTED_PROXIES`. The database path must not become a way to bypass that check for the header.
- The boot warning about ASN entries with no header only fires when neither source is available.
- `docs/content/docs/development/configuration.mdx` gains the variable in the tracking table, and the GeoIP section explains that the two databases are separate files.

## Non-goals

**This does not make the vendor ASNs safe to enable by default, and they should stay commented out.** The reason they ship commented is unchanged by this: Proofpoint Isolation and Mimecast Browser Isolation render a clicked page in the vendor's own cloud, so a whole-vendor ASN can carry a genuine human click, and isolation is typically scoped to uncategorised URLs, which is what a new outreach domain looks like. This issue is about making the mechanism reachable, not about changing which entries are on.

Context: #410, #412, #438.

meszmate:
Closes #440.

## The gap

`asn:` entries in `tracking/scanner-networks.txt` did nothing unless `TRACKING_SCANNER_ASN_HEADER` named a header a trusted proxy wrote, which in practice meant Cloudflare with a transform rule on `ip.src.asnum`. An instance behind nginx, Caddy, a cloud load balancer, or exposed directly got no ASN matching at all. That is where the mail-security vendors live, so half the catalogue was unreachable for most operators.

### 1. Native ASN resolution (`TRACKING_SCANNER_ASN_DB`)

A new `AsnDb` (`tracking/src/asndb.rs`) reads a MaxMind GeoLite2-ASN database and answers `IpAddr -> Option<u32>`. Absent, unreadable or corrupt disables ASN matching with one line at boot and never fails a request, the same rule `GEODB_PATH` follows.

- with a database and no header, `asn:` entries match on any instance, no edge configuration
- where both are set, the header wins on every request that carries one; a request it does not cover falls through to the database
- the header is still read only from a peer in `TRACKING_TRUSTED_PROXIES`. The database is a lookup on the address, so it is not a way around that check
- the boot warning fires only when neither source is usable
- pointing it at a City database is refused at boot with a line naming what it found, rather than resolving nothing and saying nothing

Two deliberate deviations from the issue text: the file is read into memory rather than memory-mapped (`open_mmap` is `unsafe` and requires the file not be replaced in place, which an operator running `geoipupdate` will do; GeoLite2-ASN is ~10 MB), and the database-type check above was added because a City file would otherwise be a silent no-op.

### 2. A `probable` certainty, and the vendor ASNs enabled

The catalogue gained a certainty column, orthogonal to scope:

- `certain` (default) the source only ever filters mail, so a match is the whole verdict
- `probable` the source can also carry a person, so a match corroborates the timing rule instead of replacing it

Proofpoint (`22843`, `26211`, `52129`), Mimecast (`30031`, `39588`, `42427`, `60492`) and Cisco (`16417`) are now **enabled and `probable`**. The flag travels on the wire as `scanner_probable`, and the consumer measures such an event against a new operator-editable window (`tracking.machine_window_probable_seconds`, default 600s, its own 1 to 86400 bounds) instead of the window for its event kind, taking whichever is wider. Outside that window the event is judged exactly as an unrecognised source would be.

That is what makes them safe to ship on. A delivery-time scan runs within minutes of the send; a recipient clicking through Proofpoint Isolation or Mimecast Browser Isolation runs whenever they got to it.

**This is a trade, not a free win.** It moves events in one direction only, from counted-as-human to counted-as-automated, so it can never let a scan count as engagement. What it costs is the recipient behind one of those vendors who genuinely clicks inside the 10-minute window: that click is now classified automated and fires nothing. The window is the dial for it. There is a second, window-independent cost: a click from any recognised network is redirected without its identification ticket, because the edge has no dispatch time, so an isolated click the consumer later counts as a person's still leaves no website-visit attribution.

The whole-cloud Microsoft and Google ASNs stay commented out even as `probable`: a recipient on Windows 365 or a corporate NAT gateway sits inside them all day, so a match there says almost nothing.

### 3. Atomic event deduplication

`is_duplicate` was `contains_key` then `insert`. Scanners fetch a message's pixel and every link in parallel, so several requests carrying one key are genuinely concurrent and every one of them saw a miss and published. Now one coalesced `entry().or_insert_with(...)`.

The race test written to prove this caught that the first fix was still racy: `or_insert` with an eager value does **not** coalesce concurrent callers (4 of 8 racers came back fresh). Only the closure form runs once per key under a lock. The reason is in the doc comment.

### 4. ASN resolution observability

A database that opens cleanly and covers none of an instance's traffic is indistinguishable from a working one. The service now counts lookups and says one thing, once: the first ASN it ever resolves, or, after 500 lookups with nothing resolved, a warning naming both likely causes (a wrong or stale file, or an unset `TRACKING_TRUSTED_PROXIES` behind a proxy). Header-resolved ASNs are not counted, so a Cloudflare instance with no database is never accused of running a silent one.

### 5. Scaling caveat documented

The tracking service's dedupe, rate limits and link caches are per-process with no shared store, so a second replica doubles a flooding source's budget and lets the same open publish twice. `configuration.mdx` now says to run a single copy and scale workers instead, with the sticky-hash escape hatch for anyone who must.

## Operator-visible changes

- new env var `TRACKING_SCANNER_ASN_DB` (compose, both env examples, the instance-config registry, `configuration.mdx`)
- new instance setting `tracking.machine_window_probable_seconds`, in the admin panel's Automated engagement card with its own bounds
- **the shipped catalogue now has live ASN entries**, so an instance with neither an ASN database nor an ASN header gets one WARN at boot saying those entries match nothing and naming both fixes. That is the honest state, and the two halves are one feature

## Verification

`cargo fmt --check`, `cargo clippy -D warnings` (default and `--features kafka --all-targets`), 43 Rust tests, `gofmt`, `make lint`, Go tests, `admin` and `docs` typecheck + lint, `docker compose config`.

Every new guard was mutation-tested: 13 mutations across Rust and Go (probable decides outright, probable never widens, the window falling below the kind window, wrong clamp bounds, the keyword ignored, everything/nothing probable, the decode field name, check-then-insert dedupe, header-over-database precedence, the empty-catalogue skip, the ASN counters) and each is caught by a named test.

The GeoLite2-ASN reader is tested against a hand-built MaxMind database (`asndb::fixture`), so there is no vendored binary and no network fetch; the third-party reader parsing it is what validates the encoder.

## CodeRabbit review

Four findings on `b7b9f2d5`, all valid, all fixed in `bb878731`:

1. **`can_resolve_asn` counted an unusable header as a source.** With `TRACKING_TRUSTED_PROXIES` empty, `peer_is_trusted` is false for every request, so a configured `TRACKING_SCANNER_ASN_HEADER` can never resolve anything, yet the boot warning was suppressed: exactly the misconfiguration the warning exists to report. `ScannerNetworks` now takes an `AsnSources` carrying whether any peer is trusted at all, and the info line says `set, but no trusted proxy to believe it from` rather than printing the header name. Covered by a test, mutation-verified.
2. **The database does not remove the `TRACKING_TRUSTED_PROXIES` requirement.** It removes the need for an ASN header and a transform rule; behind a reverse proxy without that variable the client address is the proxy's, so the lookup resolves the proxy's network. Both env examples and `configuration.mdx` now say so.
3. **A documentation claim of mine was wrong.** The catalogue and this PR both said the change "can only move scans out of your numbers, and never a click that was already counted as human." A Proofpoint click at T+2min *was* counted human before and is counted automated now, so it does reclassify clicks. The defensible statement is the one-way property above, and every surface now states the cost instead of denying it. The admin panel and the API contract comment describe classification rather than ground truth.
4. **Probable matches also suppress click identification**, independently of the window. Documented in `configuration.mdx`, `events.mdx` and the catalogue.

Context: #410, #412, #438.

## Summary by CodeRabbit

* **New Features**
  * Improved automated-engagement detection for Proofpoint, Mimecast, Cisco Secure Email, and other probable scanner networks.
  * Added a configurable probable-scanner window, allowing delivery scans to be distinguished from later human clicks.
  * Added optional GeoLite2-ASN database support for scanner-network recognition, with trusted header values taking precedence.
  * Scanner events now include whether the source is a probable scanner.

* **Documentation**
  * Updated configuration, analytics, event, and scanner-catalogue documentation with setup guidance and classification behavior.


meszmate:
@coderabbitai full review

meszmate:
@coderabbitai full review

meszmate:
@coderabbitai full review

meszmate:
@coderabbitai full review

PR #533, opened 2026-09-15
Clear the live error-tracking issues, and the workspace rename that renamed the wrong workspace

meszmate:
## What this fixes

Everything still firing in error tracking after today's v0.4.14 rollout, plus the workspace-name bug reported on Discord. The rest of the open issue list was already fixed by #519 and #532; those events are from bundles and containers that predate the deploy and decay on their own.

### Renaming one workspace renamed the other (Discord report)

`WorkspaceSettingsPage` seeded the name field from the workspace it mounted with and never re-read it, while an effect moved the autosave baseline to the newly selected workspace's name. Switching workspaces therefore left the field holding the previous name against a baseline holding the new one, autosave saw a change it had not been given, and saved the old name onto the new workspace. Creating a workspace switches to it, so the reporter's "the other changes automatically" needed no typing at all.

The editor is now keyed on the workspace id, so a switch re-seeds every field from the right workspace, and every write on the page goes through one guard that drops a save whose workspace is no longer the selected one, which also covers a debounce or a blur landing after a switch. The sending and website-tracking pages hydrate a whole settings object the same way, so they got the same pinning: their next edit after a switch used to write the previous workspace's entire settings onto the new one.

### Kafka on Confluent: one lost event and one issue per message

`ensureTopics` treated any non-`ALREADY_EXISTS` answer from `CreateTopics` as a create failure and failed the publish. A managed cluster answers `Authorization failed` for a topic that is already there, and because the topic never entered the known set, every subsequent publish retried the admin round trip and failed again: `warmup-events` alone produced 201 reports, and every warmup analytics event was dropped.

An authorization failure now reads as "not ours to create": logged once, remembered, and the produce decides. A topic that genuinely is absent fails at the produce, which says so.

### Cancelled requests are not incidents

A cancelled context reached error tracking through every layer it unwound, so today's redeploy filed nine issues naming whichever queries were in flight, and a browser navigating away did the same. Dropped centrally in `errs`, because a repository reporting a failed query cannot tell whether the request behind it still exists. `context.DeadlineExceeded` still reports: a deadline this process set and blew through is its own problem.

### Source maps were never uploaded from the Pages build

Every frontend frame in every issue reads `Invalid source map: bad json: expected value at line 1 column 1` beside a minified name. The image build uploads maps; `build:pages`, which is what a static host runs, did not, so PostHog fell back to fetching `<bundle>.js.map` from the site and got the SPA fallback. `build:pages` now runs the same upload, under the same "only when credentials are present" guard as the Dockerfile.

**This needs the two build variables set in the Cloudflare Pages projects to take effect** (`POSTHOG_CLI_API_KEY`, `POSTHOG_CLI_PROJECT_ID`); with them unset the build prints a skip line and is unchanged.

### Browser noise from the marketing site

`Script error.` (145 occurrences, 33 users) and the ResizeObserver notice were the two largest issues in tracking, and both came from warmbly.com: the dashboard and admin panel drop them, the site had `capture_exceptions: true` and no filter. Added there and on the hosted form page, which is embedded in customer sites and so sees the same opaque cross-origin errors.

### Also

`forms/src/turnstile.ts` differed from `Turnstile.tsx` only in case, which resolved both imports to one file on a case-insensitive filesystem and left `pnpm typecheck` failing in `forms/` with TS1149.

## Verification

`gofmt -l` clean, `make lint` and `golangci-lint` exit 0, `go build -tags kafka ./...`, new tests in `internal/observability/errs` and `internal/infrastructure/eventbus` (kafka-tagged) pass, `pnpm typecheck` + `pnpm lint` clean in `web/` and `forms/`, `pnpm types:check` clean in `docs/`, and the site's inline init parses.

## Summary by CodeRabbit

- **New Features**
  - Added more reliable Turnstile loading and reset behavior for forms.
  - Static page builds can automatically upload source maps when PostHog credentials are configured.

- **Bug Fixes**
  - Prevented settings drafts and saves from crossing between workspaces when switching organizations.
  - Reduced reporting of common browser-generated noise and canceled operations.
  - Improved event-bus topic handling when authorization prevents topic creation.
  - Improved handling and display of empty audit-log and admin list results.

- **Documentation**
  - Documented source-map uploads and troubleshooting for static page builds.


meszmate:
Added one more: `null is not an object (evaluating 'd.data.length')` on admin.warmbly.com/audit is the same nil-slice-to-null class as the `email_accounts` crash #519 fixed, so every list `pg_admin.go` builds is now made rather than declared (13 of them, an empty page was `null` for all of them), the audit table's own empty check is guarded, and its result type says `| null` the way `AdminUserPreview`'s slices already do.

meszmate:
Both review findings addressed in ab9612ff.

**Cached rejected promise in the Turnstile loader** — valid, and worse than it reads: a form that requires a captcha cannot be submitted at all after one blocked load, until the visitor reloads. The failure handler now clears `scriptLoading` and removes the dead tag, so the next mount makes a fresh request.

**Substring match on the broker description** — valid. That branch remembers the topic as present, so mis-classifying one error there loses every later publish to that topic with nothing reporting why. The two error codes stay the primary answer and the fallback now matches the whole normalized description (`authorization failed`, trailing period trimmed) rather than a substring, which is exactly what Confluent Cloud sends. Two negative cases added to the test: an error that merely mentions transactional-id authorization, and a SASL authentication failure.

PR #566, opened 2026-09-17
feat: add safe automatic inbox tagging and follow-up labels

rocker1166:
Adds opt-in automatic classification for inbound mail, a review page for stored decisions and confidence, and local follow-up labels derived from mailbox state. Classification is disabled unless both `TYPESAFE_API_KEY` and `INBOX_TAGGING_ENABLED=true` are set. Timestamp-only follow-up states continue to run locally without TypeSafe.

The production path now claims each message atomically before the external call, excludes workspace-owned senders, preserves low-confidence answers for review without trusting them for labels or relevance, applies labels during live ingest, and refreshes the dashboard over realtime. Follow-up state uses complete, trusted classifications in mailbox chronology, replaces only system-applied labels transactionally, preserves teammate-applied labels, and will not treat an unclassified automated message as a human reply.

The change also adds the paginated analytics endpoint and routed settings page, historical backfill and follow-up operator commands, workspace export support, configuration and data-control documentation, readable label explanations, and migration `000171`.

Validation:

- `make lint`
- targeted Go tests for inbox tagging, consumer, API handlers, repositories, instance configuration, organization transfer, consumer command, and operator CLI
- `pnpm typecheck`, `pnpm lint`, and 370 web tests
- docs typecheck and lint
- `make check-migrations` reports versions `000001` through `000171` with no duplicates

## Summary by CodeRabbit

* **New Features**
  * Added optional automatic inbox tagging for inbound messages, including labels, relevance scores, priorities, and review indicators.
  * Added an Inbox Tagging review page with filtering, summaries, pagination, confidence details, and tag explanations.
  * Added hourly follow-up labels that update as conversations progress.
  * Added operator commands for backfilling tags and recalculating follow-ups, including dry-run and resumable processing.
  * Added analytics access through `GET /analytics/inbox-tagging`.

* **Documentation**
  * Documented configuration, data controls, exports, API usage, and automatic tagging behavior.


rocker1166:
## Added: backfill over existing mail

A classifier that only sees new arrivals is no use on the day you switch it on — the inbox you want sorted is the one already sitting there. The plan's phase 1 calls for running it over the existing inbox, and this is that.

```bash
warmblyctl inbox-tag backfill --org you@example.com --days 30 --dry-run
warmblyctl inbox-tag backfill --org you@example.com --days 30 --limit 200
```

An operator command rather than a button, because a backfill over a large mailbox is real money and real time: it is bounded, resumable, interruptible, and it can be costed before it is run. Every message is idempotent on its Message-ID, so hitting the limit, Ctrl-C, or a second run all keep what was done and continue from there.

Inbound only. Own sends are excluded by folder, and so is mail from one connected mailbox to another, which lands in the second one's inbox and is still ours.

It also fills in the previous message in each thread from sent mail, so a reply reading "yes, let's do it" is judged against the thing it agrees to. The live ingest path now does the same lookup — it was previously sending replies with nothing to answer.

## Two things running it on real mail found

Both were fixed here, and both are the kind of thing only real data surfaces.

**1. A signal that lands on everything is not a filter.** The first backfill put `needs-human-judgement` on nearly every row, bounces and platform notifications included. The model answers the question honestly for any text, but "a bounce needs a person to read it" is not true. Surfaced signal labels are now gated to human replies. They are still recorded and still feed the score for every kind.

**2. An unreadable intent was discarding a readable kind.** Four messages came back `human_reply` at **0.97** confidence and were labelled `needs-review` with nothing else, because the *intent* behind them sat at 0.6. That threw the confident half away with the doubtful half.

Now the confident half stands: the kind label, the signals and the relevance score all apply, `needs-review` is added to say the intent could not be read, and the intent itself is left unset. On a 40-message sample this took `human-reply` from 3 labels to 11.

## Measured on 40 real messages

| | |
|---|---|
| classified | 40 of 40, 0 failed, 16s |
| cost | ~$0.007 per 125 messages (dry run reports it before you commit) |
| kinds | `human_reply` 11, `bounce_hard` 8, `auto_reply_ticket` 6, `notification` 2, `cold_inbound` 1 |
| categories seeded | 25 — the full taxonomy, so every label is filterable before it first fires |

**Worth a maintainer's eye:** 19 of 40 carry `needs-review`, all of them from intent confidence rather than kind. The kinds are solid; reading *what a reply wants* from these threads is genuinely harder. That is honest output rather than a bug, but if it stays that high through a week of phase 1 the intent question is what to tune, not the floor.

Verified end to end against a throwaway copy of a production database — real mail, real API calls, real category and thread-label writes — then the copy was dropped. New tests cover the backfill's dry-run, resumability, cancellation and disabled paths, plus both fixes above.

rocker1166:
## Follow-ups, tag meanings, and the self-hosted path

Rebased onto @meszmatew's review commit — the atomic claim, the instance-config entries and the address normalisation are all kept, and what follows sits on top of them.

### Follow-up labels: who owes whom a reply

| label | means |
|---|---|
| `ball-in-our-court` | They replied and we have not answered in two days |
| `awaiting-reply` | We sent last, still early |
| `follow-up-due` | We sent last, five days, silence |
| `going-cold` | They were interested, then quiet for ten days |

**No model call.** Whether a message was answered and how long ago are facts in the mailbox, and dates and arithmetic are the two things Jev is documented to be worst at. This is a SQL query and a pure function.

`going-cold` is deliberately not a longer `follow-up-due`: a stalled deal and an unanswered cold email need different things from a person, and someone who already said yes has earned more patience than someone who never answered.

Threads classified `not-interested`, `opt-out`, `wrong-person` or `not-now` are **never** chased. Nagging someone who declined is rude; nagging someone who asked to be removed is a compliance problem. Bounces and autoresponders are skipped too.

These states move with the calendar, not with any event, so they are recomputed hourly and **replaced rather than added** — `SyncExclusiveLabels` only ever touches its own family, so a label a person applied by hand is never removed. A thread wears its state, not its history. `warmblyctl inbox-tag follow-ups` forces a sweep.

On a production instance: 271 threads swept, 7 `ball-in-our-court`, 103 `follow-up-due`.

### Every tag explains itself on hover

A chip reading `going-cold` is only useful to someone who already knows the taxonomy, and nobody does on day one. `web/src/lib/unibox/tagMeanings.ts` gives each label one sentence, shown on the chip, in the scope rail, in the label menu and on the review page. The wording mirrors the criteria the classifier is actually given, so the explanation and the rule cannot drift apart. A Go test pins the label count against the dashboard's list, so a new label cannot ship without its explanation.

### The self-hosted path CONTRIBUTING asks for

> Any required third-party service needs an open-source or self-hosted path.

The tagger is now always constructed, and what it can do depends on what is configured. **With no `TYPESAFE_API_KEY` the follow-up labels still run** — they need nothing outside the instance, and "they replied and you have not" is arguably the most useful label in the set. The key adds classification, which is the only part that sends content anywhere.

So the external service is genuinely optional rather than nominally optional, and an instance without it keeps a working feature instead of a disabled one. Documented in the guide.

### Also fixed

Live-arriving mail was classified and stored but never labelled, because the consumer passed `nil` for the category store — only the backfill ever put a chip on a thread. @meszmatew had already caught this; the merge keeps their fix.

### Checks, per CONTRIBUTING

`make fmt` clean, `gofmt -l` empty, `go build`, `go vet`, `make check-migrations` clean. `pnpm typecheck` clean and `pnpm lint` 0 errors in `web/`, 41 test files / 368 tests green. Docs `types:check` clean. No Rust or Elixir touched.

rocker1166:
## Fix: `agreed` and `scheduling` were overlapping

Rebased onto @meszmatew's themed tooltips and the migration renumber.

A real reply in production came back with no intent at all:

> Hi,
>
> We are interested in that. Can we make a quick call about it?

```
agreed:     0.52
scheduling: 0.46     confidence 0.46 -> under the floor -> no intent, needs-review
```

The model was right to hesitate. "We are interested" is agreement and "can we make a quick call" is a request to talk, and the criteria as written made both true at once. So the single most valuable reply shape in cold outreach was the one that reliably got no intent.

The boundary is now stated explicitly rather than left implicit, which is what the jaggedness guidance asks for:

- `agreed` — "…Use this when they say yes or ask to talk **without naming a specific time**"
- `scheduling` — "Names an **exact time, calendar date, or day of the week**… A vague period such as 'next week' is not exact enough; use agreed for those."

Measured on seven shapes, before and after:

| message | before | after |
|---|---|---|
| "interested, can we make a quick call?" | scheduling 0.62 **under floor** | agreed **1.00** |
| "let's set up a call sometime next week" | scheduling 0.57 **under floor** | agreed **1.00** |
| "can we jump on a call next week + calendar link" | scheduling 0.99 | agreed 0.88 |
| "would 2pm GMT+8 on 22nd Sep work?" | scheduling 1.00 | scheduling 1.00 |
| "could we talk on Thursday?" | scheduling 1.00 | scheduling 1.00 |
| "Tuesday 3pm works, sending an invite" | scheduling 0.99 | scheduling 1.00 |
| "yes, let's do it" | agreed 1.00 | agreed 1.00 |

Two floor breaches removed, no regression, and the ones that name a real time still land on `scheduling`.

In production that message went from `needs-review` at relevance 34 to **`agreed` at 1.00, relevance 84, priority `now`** — from buried to the top of the inbox, which is where "we are interested, can we talk" belongs.

## Re-running verdicts a criteria change rescues

Changing the weights can be re-applied to stored answers for free. Changing what the model is **asked** cannot: the stored answer is an answer to a question that no longer exists. `--redo-unreadable` clears the under-floor verdicts and their `needs-review` labels so the next backfill asks again, bounded to those rows because a confident answer needs no rescuing.

```bash
warmblyctl inbox-tag backfill --org you@example.com --redo-unreadable
```

## One fixture recording bug worth noting

The fixture recorder was pinning the API's raw choice rather than what `Decide()` produces, so a fixture could claim a label the system drops under the floor. It now records the system's behaviour, which is the thing the tests exist to pin.

## Still open

36 of 125 carry `needs-review`. 11 are intent on a confident human reply; the rest is the model unsure what the message **is**, dominated by `cold_inbound` vs `notification` (13 between them). Those two genuinely overlap — a "we've launched X" mail from a vendor you already know is arguably both — so that is the next criteria boundary to draw, not a floor to lower.

`make fmt`, `gofmt -l`, `go build`, `go vet`, `make check-migrations` clean.

Issue #575, opened 2026-09-17
Warmup pairs an organization's own mailboxes together: 61% of received warmup mail came from our own 20 mailboxes


ClosiQode:
## Summary

An organization with 20 mailboxes in the premium pool warms overwhelmingly against itself. Partner selection excludes only the sending mailbox, so the sender's 19 sibling mailboxes are candidates on every draw, and with a thin premium tier they win most of them.

Measured on a self-hosted instance linked to Warmbly Cloud, 20 `smtp_imap` mailboxes across 8 domains and 3 providers, all `premium`, day 0 to day 1:

| | |
|---|---|
| Warmup messages received, all mailboxes | 64 |
| From the organization's own mailboxes | 39 (61%) |
| From external partners | 25 |
| Distinct external partners | 22, across 15 domains |

In the densest window (20:17-20:29 UTC), 15 consecutive arrivals were all internal, sender and recipient both belonging to this organization. Verified from the receiving side: the messages carry `X-Mailtrace-Verify` and were submitted from the pool's own infrastructure (`static.*.clients.your-server.de`, `2a01:4f8::/29`), so this is the cloud choosing these pairs, not local warmup.

## Why

`WarmupPartnerCandidates` (`repository/pg_warmup.go:953`) filters on pool type, participant role, account status and health, and excludes exactly one mailbox:

```sql
AND wpp.email_account_id <> $2
```

Nothing excludes the same organization or the same owner. The borrow path only widens the set when the own tier is thin (`WarmupPoolTierFallbackFloor = 25`), and borrowed partners must additionally be healthy, never blocked, and a member for `WarmupPoolFallbackMinAgeDays`, which a recently joined mailbox fails. A customer who brings 20 mailboxes therefore makes the problem worse, not better: the more mailboxes they enroll, the larger their share of their own candidate set.

`resources/warmup-abuse-cold-research.md` and issue #150 already describe this shape for small pools ("diversity collapses into near-uniform reciprocal reuse"). The same collapse happens at the premium tier once one organization is a large fraction of it.

## Why it matters

Warmup exists to build reputation with the providers that will receive the cold mail, which for us means Google and Microsoft. Mail that a customer's own mailboxes exchange between their own domains, on their own providers, teaches those providers nothing and costs the customer their daily ramp. It is also the traffic pattern a spam filter is most likely to read as a closed loop.

## Suggested

[...]

PR #578, opened 2026-09-18
feat: fix the four standing defects in #574 and the same-workspace warmup pairing in #575

### LINKED ISSUE #574: Cloud warmup on SMTP/IMAP mailboxes: nothing is delivered (SERVER_UNREACHABLE), yet the day's counter keeps climbing past the target
## Summary

Self-hosted v0.4.22 linked to Warmbly Cloud. 20 mailboxes, all `smtp_imap`, enrolled in the pool across three providers (Infomaniak, o2switch/cPanel, LWS). `warmup_base=10`, `increase=1`, `max=40`, day 0.

Two things go wrong together:

1. **No warmup mail actually leaves these mailboxes.** The provider's own delivery log shows zero outbound submission, and several mailboxes carry `SERVER_UNREACHABLE`.
2. **The day's counter keeps growing anyway**, to 14-21 against a target of 10, because a failed warmup send is never given back.

The same instance warmed correctly until 16/09 while its mailboxes were Google Workspace (OAuth). The only change since is the mailbox type: OAuth -> SMTP/IMAP.

## Evidence

**The provider says nothing was submitted.** On the o2switch/cPanel account hosting 5 of the mailboxes, API2 `EmailTrack::search` (Exim delivery log, 659 rows since the previous morning, `max_results_by_type=5000`) returns **zero authenticated outbound message** from any enrolled mailbox, for any of `deliverytype=remote|all` or a per-sender query. Other authenticated submissions from the same account on the same day are logged normally, so the log is not the problem.

**Inbound warmup works.** Each mailbox has 1 to 5 partner messages filed in `INBOX.Warmbly`. Reading that folder on `mailbox-a@domain-1.example` (19 sends reported today) shows 5 messages, **all of them new threads opened by partners, not one `Re:`**. With `reply_rate=30` and 19 outgoing threads, replies should be visible.

**Credentials, ports and TLS are fine.** From a French IP, authenticating against each provider succeeds in under 1.5 s, on both ports, with the exact credentials stored in the pool:

| Mailbox host | 465 implicit TLS | 587 STARTTLS |
|---|---|---|
| kilo.o2switch.net | AUTH OK | AUTH OK |
| mail97.lwspanel.com | AUTH OK | AUTH OK |
| mail.infomaniak.com | AUTH OK | AUTH OK |

**The failures span all three providers**, which points at the sending side rather than at one host. Note that IMAP from the self-hosted worker (France) works throughout: the `Warmbly` folder is created and filled. Only the cloud's SMTP submission fails.

## Where it breaks, exactly

`SERVER_UNREACHABLE` carrying the canned text "The connection to the mail server could not be established" is `errx.ErrMailServerUnreachable`. In `client/smtpimap/smtp/client.go` it is returned from eleven distinct points (lines 293, 306, 314, 327, 330, 363, 374, 396, 410, 417, 431) as a bare sentinel, with the underlying error dropped. A dial refusal, a DNS failure, an EHLO rejection and a STARTTLS failure are therefore indistinguishable from the outside, and from your own logs unless the raw error is captured before it is replaced.

The first of those, line 293, is the dial:

```go
if implicitTLS {
    conn, err = netbind.TLSDialer(c.BindIP, tlsConf).DialContext(ctx, "tcp", addr)
} else {
    conn, err = netbind.Dialer(c.BindIP).DialContext(ctx, "tcp", addr)
}
if err != nil {
    return errx.ErrMailServerUnreachable
}
```

`netbind.Dialer` sets `LocalAddr` from the client's `BindIP`, falling back to `WORKER_BIND_IP`. The package comment states the intent: "Hosted Warmbly on a multi-IP box: WORKER_BIND_IP set per process so SMTP/IMAP/HTTPS egress leaves from the assigned IP", and `bind_ip` is handed to each worker by `/internal/worker-config` (`api/handler/internal_worker_config.go`).

**`netbind` is used by the SMTP and IMAP clients only.** Grepping `netbind.` across `internal/` returns `client/smtpimap/smtp/client.go`, `client/smtpimap/imap/client.go` and `internal/email/*`; neither `client/goog` nor `client/msgraph` appears. OAuth mailboxes therefore egress over HTTPS on the default route, while `smtp_imap` mailboxes have their outbound TCP pinned to a source address.

That is exactly the asymmetry observed here: through the same pool, on the same day, Google Workspace mailboxes warmed normally and `smtp_imap` mailboxes reach no provider at all. A `bind_ip` that is not assigned on the host makes `DialContext` fail immediately with `cannot assign requested address`, which surfaces as this error for every SMTP mailbox regardless of destination. Blocked egress on 465/587 from the sending infrastructure would look identical.

Worth checking on the pool side, in this order: the `bind_ip` served to the worker that hosts these mailboxes and whether it is actually configured on that host; whether outbound 465 and 587 are permitted from it; and the raw dial error before line 293 replaces it.

## Why the counter still climbs

`tasks/email_task.go` stamps the day before the worker has done anything:

1. STEP 10 `emailSender.Send` only checks worker liveness and publishes `SendEmailParams` (`email_sender.go`, `PublishSendEmail`). It returns `nil` as soon as the event is queued.
2. STEP 12 `warmupRepo.IncrementDailyCount` -> `warmup_statistics.emails_sent += 1`.
3. STEP 13 marks the task `completed`.

When the worker then fails the real send, `consumer/event_send_result.go` `HandleEmailFailed` sets the task to `failed` and then:

```go
switch task.TaskType {
case "campaign":
    return s.failCampaignSend(ctx, task, reason, code, nil)   // "the day's counters give the send back"
case "email":
    s.notifyUserSendFailed(ctx, task, reason)
}
```

There is no `case "warmup"`, and nothing decrements `warmup_statistics.emails_sent` anywhere (the only give-backs found are the campaign ones, `pg_campaign_progress.go:467` and `pg_campaign.go:2116`).

The cap and the displayed number then read different sources:

| Number | Source |
|---|---|
| Cap enforced by the scheduler (`warmup_scheduler.go` STEP 3/4) | `CountWarmupEmailsSentToday` = `COUNT(*) FROM tasks WHERE status='completed' AND task_type='warmup' AND DATE(completed_at)=CURRENT_DATE` |
| Number displayed, and sent to the linked instance | `analytics.GetAccountDailyUsage` = `COALESCE(ws.emails_sent,0) FROM warmup_statistics`, surfaced as `DailyUsage.WarmupSent` and copied to `PoolLinkMailboxState.SentToday` (`poollink/service.go:571`) |

A failed attempt leaves `tasks.status='failed'`, so it stops counting against the cap and the scheduler immediately allows another send, while `warmup_statistics.emails_sent` keeps it forever. The loop is: dispatch, count, fail, uncount the cap, dispatch again. Same defect class as #312, fixed for campaigns and not for warmup.

## Expected

- Cloud warmup on `smtp_imap` mailboxes actually delivers, or the mailbox is surfaced as failing instead of appearing to warm.
- `HandleEmailFailed` gets a `case "warmup"` that gives the day back (`emails_sent`, and `emails_replied` when the attempt was a reply).
- The cap and the reported number read the same source, so `sent_today` cannot exceed `target_volume`.

## Environment

- Warmbly v0.4.22, Docker Compose, Postgres 16, Redis 7, NATS 2.10
- 20 `smtp_imap` mailboxes, all enrolled in Warmbly Cloud; local warmup idle (`warmup_statistics` and `warmup_tokens` empty on the instance, 20 rows in `cloud_link_mailboxes`)
- SMTP 465 implicit TLS, IMAP 993
- Same instance warmed correctly through the cloud until 16/09 with Google Workspace (OAuth) mailboxes

Happy to run any check on the instance, or to provide the raw cPanel delivery log.

meszmate:
Fixes #574 and #575, both reported by @ClosiQode against v0.4.22 (self-hosted, linked to Warmbly Cloud, 20 `smtp_imap` mailboxes).

Every claim was checked against the code first. All of them held up, except the `netbind`/`bind_ip` hypothesis in the original body of #574, which the reporter had already retracted himself.

## #574

**1. `SERVER_UNREACHABLE` could not be diagnosed.** `errx.ErrMailServerUnreachable` was returned as a bare sentinel from twelve distinct points in `internal/client/smtpimap/smtp/client.go`, with the underlying error dropped, so a refused dial, a timeout, an EHLO rejection and a TLS handshake failure were indistinguishable. There is now `errx.ErrMailServerUnreachableAt(stage, cause)`, which keeps the `SERVER_UNREACHABLE` code every caller switches on and names the step plus the cause:

```
The connection to the mail server could not be established
  (dial smtp.example.com:465: connect: connection refused). ...
  (dial smtp.example.com:465: tls: first record does not look like a TLS handshake). ...
  (mail from: 451 "4.7.1 try again later"). ...
  (starttls: the server does not offer STARTTLS). ...
```

`dialCause` drops the addresses `net.OpError` prints around the syscall failure: the local one is the worker's own egress IP and is not the mailbox owner's to read, and the remote one is already in the stage.

**2. `465` is accepted everywhere but did not work through the pool.** The repo's own path is right (`ResolveSMTPSecurity` maps `465` to implicit TLS and dials TLS-first), so what was missing was not the behaviour but the ability to see it. With the message above, the two cases are now distinguishable from the mailbox drawer alone: a refused or timed-out `dial` on `465` while `587` works is a network that does not allow implicit TLS submission. Documented in the mailboxes guide and as its own row in the self-hosting troubleshooting table.

**3. Deleting a mailbox never unenrolled it from the cloud.** `emailService.Delete` removed the mailbox from local warmup pools and let the `cloud_link_mailboxes` row go by `ON DELETE CASCADE`, so the pool kept the mailbox, its stored SMTP/IMAP credentials and its pool membership. The delete now revokes the enrollment first, before anything local moves, and refuses rather than deleting half-way when the cloud cannot be reached (the same posture the worker removal already takes, and for the same reason: afterwards there is no row left to retry from). A cloud-managed mirror is skipped, because `cloudlink` deletes those through this very path and calling back would recurse.

**4. A failed warmup send was never given back.** `HandleEmailFailed` had no `case "warmup"`, so a send the worker refused stopped counting against the scheduler's cap (completed warmup tasks) while `warmup_statistics.emails_sent` kept it forever: dispatch, count, fail, uncount the cap, dispatch again. Two changes:

- `FailWarmupSend` records the task failure and reverses `emails_sent` (plus `emails_replied` for a reply) in one transaction. A database failure leaves the task retryable, duplicate delivery is idempotent, counters are floored at zero, and the refund stays scoped to the day the send was counted on.
- `GetAccountDailyUsage` now counts completed warmup tasks for `warmup_sent`, exactly as it already did for `campaign_sent`, so the reported number and the enforced cap are one source and `sent_today` cannot read above the target.

## #575

`WarmupPartnerCandidates` excluded only the sending mailbox, so a workspace's other 19 mailboxes were candidates on every draw and won 61% of them.

Each candidate now carries its owner, and the selector's bucket order puts the sender's own workspace last: every available outside partner is used before any sibling. It is a preference, not an exclusion, because a self-hosted instance's pool *is* one workspace, and excluding siblings in SQL would stop warmup entirely there. An unattributable mailbox (no organization) is never read as a match. A tick with no outside partner left says so in the logs, at info, since on an unlinked self-hosted instance that is the normal case.

The diversity a mailbox is getting is also surfaced from confirmed deliveries only: `partner_mailboxes_7d`, `partner_domains_7d` and `partner_organizations_7d` on `warmup_health`, shown next to local warmup reputation or inside the Warmbly Cloud card for linked self-hosts, with a note when a whole week.s partners were in one workspace.

**Not in this PR:** the routing-rules UI the issue asks for as a third suggestion. It existed as a mitigation for the pairing bug (the reporter created 64 rules by API to work around it), and with same-workspace pairing fixed by default it is no longer the only lever a customer has. Filed as a follow-up rather than bundled into a bug-fix PR.

## Verification

No migration, so no schema change and no numbering to race.

- `make fmt`, `make lint` (includes `check-migrations`, `check-join-script`, the installer and dockerfile checks) clean
- `pnpm typecheck` + `pnpm lint` clean in `web/`; `pnpm types:check` + `pnpm lint` clean in `docs/`
- new tests, each one mutation-tested by reverting the fix and watching it fail:
  - `internal/client/smtpimap/smtp/unreachable_detail_test.go`: the stage and the server's own reply survive for a refused dial, a missing STARTTLS, an implicit-TLS handshake, and 4xx at auth / mail from / rcpt to / message accept; `dialCause` keeps the failure and drops both addresses
  - `internal/tasks/partner_selection_org_test.go`: an outside partner wins over 19 siblings across 20 draws and over a fresh sibling even when recently used; a single-workspace pool still warms; an ownerless mailbox counts as outside
  - `web/src/components/app/emails/WarmupPartnerDiversity.test.tsx`: local and cloud-linked cards render confirmed seven-day diversity and the same-workspace warning, while older cloud responses remain compatible
  - `internal/repository/warmup_daily_usage_live_test.go`: `warmup_sent` counts the delivered send and not the failed one, is scoped to type and day, and `FailWarmupSend` atomically records the failure and reverses sent (and replied only for a reply), retries database failures, is idempotent, and leaves other days alone
  - `internal/repository/warmup_partner_org_live_test.go`: candidates carry their owner; `GetPartnerDiversity` counts distinct mailboxes, domains and workspaces in the window
  - `internal/app/email/cloud_unenroll_delete_test.go`: the revocation happens before the worker removal and the row delete, a refusal or an unreadable enrollment keeps the mailbox, a managed mirror does not recurse, an unenrolled mailbox makes no round trip
  - `internal/app/consumer/warmup_send_giveback_live_test.go`: end to end through `HandleEmailFailed`, including that three deliveries of one failure give one send back, and that the cap and the reported number then agree
- full `internal/repository`, `internal/app/consumer`, `internal/app/analytics`, `internal/app/warmup`, `internal/app/email`, `internal/tasks`, `internal/client/...` live suites against a freshly migrated scratch database: green except `TestLiveRepairedVerificationRestoresRouting`, which fails identically on an unmodified `main` worktree (checked)

## Summary by CodeRabbit

- **New Features**
  - Warmup now prioritizes partners from other workspaces when available.
  - Account analytics and the Warmup view show seven-day partner mailbox, domain, and workspace diversity.
  - Failed warmup sends no longer permanently consume daily sending capacity.

- **Bug Fixes**
  - Deleting an enrolled mailbox now revokes its cloud enrollment first and preserves the mailbox if revocation fails.
  - SMTP connection errors identify the failed connection stage, improving troubleshooting.
  - Retryable SMTP failures are no longer recorded as permanent recipient evidence.

- **Documentation**
  - Updated API references, troubleshooting guidance, and Warmbly Cloud mailbox deletion behavior.


meszmate:
@coderabbitai full review

PR #579, opened 2026-09-18
Hostinger's DKIM selectors are hyphenated, not numbered

joaoppa:
`providerSelectors` maps `hostinger` to `hostingermail1` / `hostingermail2`. Those answer on no domain I have been able to test; the pair Hostinger actually publishes is `hostingermail-a` / `hostingermail-b`.

Measured across five Hostinger-hosted domains on one instance, all with `mx1.hostinger.com` / `mx2.hostinger.com` so the provider match itself is working:

```
domain                 hostingermail-a   hostingermail1
rakhiapp.com.br              found            -
usarakhi.com.br              found            -
rakhiescolas.com.br          found            -
altius.net.br                found            -
pedelimao.com.br             found            -
```

The effect is the one #528 set out to remove: the MX lookup correctly identifies Hostinger, the selector probe then misses, and every Hostinger-hosted sender is reported without DKIM even though the key is published and signing. On this instance that was 26 of 29 mailboxes.

Both forms are kept rather than swapped. I can only speak for the domains above, and a provider serving different plans or regions from different selector names is exactly the case this list exists to absorb; the extra lookup costs one DNS query on a path that already tries several.

gofmt, `go build ./...`, `go vet` and golangci-lint (v1.64.8) clean; `go test -race -count=2 ./internal/pkg/dnsauth/` passes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Hostinger email authentication detection by recognizing additional selector naming formats.


meszmate:
@coderabbitai full review

PR #584, opened 2026-09-18
Release a cloud-managed mirror's cloud link when the mailbox is deleted (#582)

### LINKED ISSUE #582: Deleting a cloud-managed mailbox from Mailboxes leaves it linked to the instance with no mirror, which the code itself calls unrecoverable
Found while fixing #574, and pre-existing: #578 does not change this behaviour, it only made the surrounding path explicit enough to notice.

## The state

A cloud-managed mailbox (signed in through Warmbly Cloud, tokens brokered from it) has a credential-free mirror on the instance. Two paths remove that mirror:

- **Settings > Warmbly Cloud > unenroll** calls `cloudlink.removeManaged` (`internal/app/cloudlink/managed.go:196`), which deletes the cloud's `instance/mailboxes/<remote>` link first and then the mirror. The mailbox stays in the cloud workspace and becomes adoptable again.
- **Mailboxes > delete** calls `emailService.Delete`, which never touches the cloud for a managed mailbox. The mirror and its `cloud_link_mailboxes` row go, and the cloud keeps the mailbox *still linked to this instance*.

`mirror()` states what that second outcome costs, in its own rollback comment (`managed.go:110`):

> a mailbox left linked to this instance with no mirror here is hidden from the adoptable list and refused on a second attempt, so it could never be recovered from either side.

Which is exactly the state a direct delete produces.

## Why #578 did not fix it

#578 added a cloud revocation to the delete path, for enrolled `smtp_imap` mailboxes whose SMTP/IMAP password the pool holds. It deliberately skips managed mailboxes, because `cloudlink.removeManaged` deletes the mirror *through* `emailService.Delete`, and calling back into the cloud from there would recurse until the stack ran out. The guard is `link.Managed`, pinned by `TestDeleteDoesNotCallBackForAManagedMailbox`.

## The shape of a fix

The recursion is now avoidable, which it was not before: `cloudlink.RevokeForDelete` (added in #578) releases the instance's claim on the cloud and drops the local row without calling the email service at all. Letting the delete path use it for managed mailboxes too would release the claim and leave the mailbox adoptable, and a second call is answered `pool_link_mailbox_not_found`, so `removeManaged`'s own flow still works.

What needs deciding first is the product question, not the code: deleting a managed mirror should probably mean "give it back to the cloud workspace" (release the claim, stay adoptable) rather than "abandon it" (claim held, adoptable by nobody). The guides already say the mailbox stays in the cloud workspace when removed from the instance, which reads as the first of the two.

Worth checking alongside it whether any existing instance has already produced such a mailbox, since recovery needs the cloud side to release the claim.

tunglambk:
Fixes #582.

Deleting a cloud-managed mailbox from **Mailboxes** now releases the instance's claim on the cloud before the local mirror goes, so the mailbox returns to the cloud workspace and can be adopted again. Before this, the mirror and its `cloud_link_mailboxes` row disappeared while the cloud still believed the mailbox was linked here. That is the state `mirror()`'s rollback comment calls unrecoverable: hidden from the adoptable list, refused on a second attempt.

The fix is the one the issue outlines. `cloudlink.RevokeForDelete` already releases the cloud side and drops the local row without touching the email service, so `unenrollFromCloud` no longer skips `link.Managed` and calls it for every cloud link. `removeManaged` still deletes the mirror through `emailService.Delete`, so its flow makes a second cloud call; the cloud answers that one `pool_link_mailbox_not_found`, which `RevokeForDelete` tolerates, so the Settings path keeps working.

No recursion. `RevokeForDelete` is a leaf, `TestDeleteReleasesTheCloudLinkForAManagedMailbox` pins that the delete entry point runs exactly once for a managed mirror, and `TestRevokeForDeleteDoesNotCallIntoTheMailboxDelete` pins that the revocation never reaches the mailbox service. `TestUnenrollReleasesTheCloudBeforeDeletingAManagedMirror` covers the two-call Settings flow.

The 409 text in `ErrCloudEnrollmentStuck` and its docs now say the link could not be released instead of the password being left in the pool, since the delete path can return it for a managed mirror too.

## Verification

- Reproduced first: the new managed-delete test fails on unmodified `main` with `revoked [], want one call for <id>`.
- `go build ./...`, `go vet` on both touched packages, and `gofmt -l .` clean.
- golangci-lint v1.64.8 (the version CI pins) clean on both touched packages. It has to run under `GOTOOLCHAIN=go1.25.0`; the local Go is 1.27.1 and v1.64.8 cannot read its export data.
- `go test ./internal/app/email/ ./internal/app/cloudlink/` green. Both suites are unit tests over stubs and `httptest` servers, so no database or live cloud is involved.
- `pnpm types:check` and `pnpm lint` clean in `docs/`; the two eslint warnings are pre-existing and unrelated.
- Not in this PR: recovering a mailbox that an earlier direct delete already orphaned on the cloud side. That needs the cloud to release the claim, which is the API-side half the issue mentions.

## Summary by CodeRabbit

- **Bug Fixes**
  - Deleting a mailbox now releases its Warmbly Cloud link before removing the local mailbox.
  - Cloud-managed mailbox mirrors are released back to the cloud workspace, allowing them to be adopted again.
  - Improved handling prevents recursive deletion behavior and ensures cloud release failures restore the mailbox appropriately.

- **Documentation**
  - Updated mailbox deletion and error-code documentation to clarify cloud link release, credential handling, mirror behavior, and related failure conditions.


meszmate:
@coderabbitai full review

PR #585, opened 2026-09-18
fix: an instant conditional branch no longer applies its target step's wait_after (#583)

### LINKED ISSUE #583: # Instant conditional branch still honors target step `wait_after`, causing unexpected delay
# Instant conditional branch still honors target step `wait_after`, causing unexpected delay

## Environment

- Warmbly version: `0.4.22`
- Deployment: Self-hosted
- Deployment method: Docker Compose
- Campaign timezone: `Asia/Kolkata`

## Summary

An instant conditional branch is not being executed immediately.

The branch is configured with:

```json
"instant": true
```

and the campaign builder visually shows the branch as instant.

However, the target sequence still has:

```text
wait_after = 10
```

and Warmbly applies that delay at runtime.

This results in a lead waiting 10 days even though no 10-day delay is visible in the campaign builder.

---

## Campaign

Campaign ID:

```text
b07e3016-715f-4244-b305-b653f409f000
```

Affected contact:

```text
shilxxxxxx@gmail.com
```

Contact ID:

```text
cafd4886-b7d1-47a9-b49f-8151b58b8e9d
```

---

## Campaign flow

Step 1 contains an open condition:

```json
{
  "instant": true,
  "branch_id": "0200315c-c181-4560-8c96-3f5671b753a5",
  "conditions": [
    {
      "field": "opened",
      "value": 1,
      "operator": "within_days"
    }
  ],
  "target_step_id": "216e4957-8e36-45ac-a098-b9892707abf8"
}
```

The condition means:

```text
IF opened the email within 1 day
```

The target is Email 3:

```text
216e4957-8e36-45ac-a098-b9892707abf8
```

The campaign builder visually shows this branch as instant.

---

## Database state

The Step 1 branch points to Email 3 with:

```text
instant = true
```

However, Email 3 currently contains:

```text
id         = 216e4957-8e36-45ac-a098-b9892707abf8
position   = 3
wait_after = 10
```

Other steps in the same campaign contain:

```text
Step 1   wait_after = 0
Email 2  wait_after = 1
Email 3  wait_after = 10
Email 4  wait_after = 1
```

There is no visible 10-day delay on the Email 3 branch in the campaign builder.

---

## Runtime evidence

Step 1 was successfully sent and completed.

```text
sequence_id  = 0c0dea46-95cc-4145-8625-3217ac6c7da8
status       = completed
scheduled_at = 2026-09-17 09:40:59+00
completed_at = 2026-09-17 09:40:59.978464+00
```

The contact then opened the email.

`campaign_contact_progress` shows:

```text
opened_at      = 2026-09-17 11:04:08.620043+00
instant_fired  = {open}
opened_machine = false
```

This confirms that:

- the open event was detected
- the open was not classified as a machine open
- the instant open condition fired successfully

---

## Actual behavior

Despite `instant_fired = {open}`, the contact activity shows:

```text
NEXT: Email 3
Waiting
not before Sep 28, 2026, 08:00 AM
Waiting 10 days after Step 1
```

This exactly matches the persisted:

```text
wait_after = 10
```

on Email 3.

---

## Date calculation

Step 1 completed on:

```text
2026-09-17
```

Applying a 10-day delay results in:

```text
2026-09-27
```

Sep 27, 2026 is outside the configured sending window, so Warmbly moves the send to the next valid campaign window:

```text
Sep 28, 2026, 08:00 AM
```

This matches the UI exactly.

---

## Expected behavior

When a conditional branch is configured with:

```text
instant = true
```

the target step should execute at the next eligible sending opportunity without applying the target sequence's normal `wait_after`.

For example:

```text
open detected
-> instant condition fires
-> target step selected
-> target step scheduled immediately / at next valid sending window
```

The target step's regular delay should not override an explicitly instant conditional branch.

---

## Actual execution behavior

The observed runtime behavior appears to be:

```text
open detected
-> instant condition fires
-> target step selected
-> target sequence wait_after is read
-> wait_after = 10 is applied
-> Email 3 delayed by 10 days
```

So the `instant` flag is successfully detected and recorded, but it does not override the target sequence delay.

---

## Possible root cause

There appear to be two possible related problems.

### 1. Runtime handling issue

The campaign engine may select the target step correctly for an instant branch, but then still apply:

```text
target_sequence.wait_after
```

when calculating the next execution time.

In that case, `instant=true` is effectively ignored for scheduling.

### 2. Campaign builder / persistence issue

It is also possible that Email 3 previously had a 10-day delay, and after configuring the conditional branch as instant, the campaign builder updated:

```json
"instant": true
```

but did not reset the target sequence's persisted:

```text
wait_after
```

This would leave the campaign in an inconsistent state:

```text
branch = instant
target step wait_after = 10
```

The campaign builder does not expose this contradiction.

---

## Additional observation

`campaign_logs` does not appear to log the instant conditional transition.

The only matching campaign log entry for this contact is the initial email send:

```text
event_type = email_sent
message    = Email sent to shilxxxxxx@gmail.com
```

No campaign log entry was created around the time the instant open condition fired.

However, the condition firing is clearly recorded in:

```text
campaign_contact_progress.instant_fired = {open}
```

Logging the branch transition would make troubleshooting this type of issue much easier.

---

## Expected consistency

The following three layers should agree:

```text
Campaign builder
Database campaign definition
Runtime lead activity
```

Currently they do not.

The campaign builder shows:

```text
instant
```

The branch metadata contains:

```text
instant = true
```

But runtime applies:

```text
wait_after = 10
```

from the target sequence.

---

## Suggested expected logic

Conceptually, scheduling could behave like:

```text
if branch.instant == true:
    effective_wait = 0
else:
    effective_wait = target_sequence.wait_after
```

or the campaign editor could ensure that an instant branch cannot retain an incompatible target-step delay.

---

## Impact

This can cause campaign emails to be delayed by days even though the visual campaign flow shows an instant conditional branch.

For users, this is difficult to diagnose because the unexpected delay is not visible in the campaign builder.

It can affect:

- campaign timing
- follow-up sequences
- behavioral/open-based branching
- time-sensitive campaigns
- testing and campaign validation

---

## Screenshots available

Image

Image

Screenshots can be provided showing:

1. Campaign flow with the `opened within 1 day` condition
2. Instant branch leading to Email 3
3. No visible 10-day delay on that branch
4. Lead activity showing:
   - `Waiting 10 days after Step 1`
   - `not before Sep 28, 2026, 08:00 AM`
5. Campaign sending schedule

tunglambk:
Closes #583.

An instant branch routes on a signal that already happened, so once it matches its target step is due now. It wasn't: `route()` derived the target's due time from that step's own `wait_after`, and `placeCampaignSend` re-derived the same wait in STEP 4. Both looked at the step and neither looked at the branch, so `"instant": true` pointing at a step with `wait_after = 10` still waited ten days while the canvas drew an instant edge.

The branch's own `within_days` window still decides when the condition is evaluated, which is the window the guide says instant respects. What changes is that once it matches, the target is due immediately rather than a further `wait_after` later.

I fixed it at runtime rather than in the branch writer. A step isn't owned by one branch: Email 3 can be the target of an instant open branch and of the cold `wait 3 days` path at the same time, so rewriting the target's stored `wait_after` would change the other path and every campaign already in the database. The branch's flag wins over the target step's own `wait_after`, and that rule now lives in one place, `repository.BranchWaitAfter(targetWaitAfter, instant)`, used both by routing's due time and by the placer so placement can't add the delay back. The pair carries `Instant` (routing to `ContactSequencePair` to `placeCampaignSend`), and the contact preview passes it through, so lead activity and the contact state stop reporting a wait that isn't there.

`nil` means instant, the same default `MatchInstantBranchTarget` and the canvas's `isInstantBranch` already use, so a branch with no `instant` field stops disagreeing with the badge it is drawn with. An explicit `false` keeps the target's wait, and a branch whose conditions aren't an instant-capable signal (`opened`, `clicked`, `reply_*`) keeps it too, even if something stamps `instant: true` on it. No schema, builder, or stored-data change.

[The sequences guide](https://github.com/warmbly/warmbly/blob/main/docs/content/docs/guides/sequences.mdx) already claims this ("Instant branches skip the wait"), so there is no docs change here.

Validation:

- `go test ./internal/repository/ -run TestRoutedWaitHonorsTheBranchInstantFlag` — new unit test over the route decision itself. Fails on `main` for both instant cases (`DueAt` ten days out) and passes with the change; the opted-out, `wait_after = 0`, and non-instant-capable cases pin the boundaries.
- `go test ./internal/repository/ -run TestLiveFindRoutedPairsInstantBranchTargetIgnoresTheStepWait` — new live Postgres test over `FindRoutedPairs`. Asserts the returned pair is due now and marked `Instant`, then flips only the branch flag to `false` and asserts the target's ten-day wait comes back.
- `go build ./...`, `go vet ./internal/repository/... ./internal/scheduler/... ./internal/app/contact/...`, `gofmt -l .` — clean.
- `go test ./internal/repository/... ./internal/scheduler/... ./internal/app/contact/...` against a migrated local Postgres (`WARMBLY_TEST_DB`) — pass.
- `make lint` (golangci-lint v1.64.8, `GOTOOLCHAIN=go1.25.0`) — passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Instant branches now route contacts to their next campaign step without applying an additional wait.
  * Branches that are not instant, or explicitly opt out, continue honoring the configured step delay.
  * Campaign scheduling and contact previews now consistently reflect instant branch timing.


meszmate:
@coderabbitai full review

PR #609, opened 2026-09-19
feat: suppression add dialog, row menu fix, and warmup suppression cleanup

meszmate:
## Summary

- **Add dialog.** Adding to the suppression list now opens a modal instead of an inline form. Pasted values are previewed as chips (addresses, domains, and unrecognized values struck through) with counts, and the primary button reads "Add N". Escape and click-outside close it, a dirty draft confirms before discarding, and Ctrl/Cmd + Enter submits.
- **Row menu.** "Remove from list" now uses the menu item's `icon` prop, so the icon sits beside the label instead of above it. The added date no longer wraps. Danger menu items across the app keep their icon red instead of grey.
- **Migration 000186.** Removes suppression entries whose only bounce or complaint evidence came from a warmup send, and removes the warmup bounce and complaint events from the deliverability record. An address that a campaign send also bounced is kept. The bounce and complaint paths already stop at a warmup task, so this only cleans up existing rows.

## Checks

- `pnpm typecheck` and `pnpm lint` in `web/`: no errors, no warnings in touched files
- `make check-migrations`: 000186 is next after `main`
- Migration applied cleanly on a fresh seeded database; dialog and menu checked in the running dashboard


meszmate:
@coderabbitai full review

PR #612, opened 2026-09-19
feat: add a Today's sending plan to the campaign overview and feed the sidebar meter and wizard estimate from the scheduler's own clamps (issue #606)

### LINKED ISSUE #606: Better view on how many email are sent per day in a campaign
Even though users set 50 emails/mailbox, it seems there's more mechanisms behind the scene that are affecting how many would be delivered and it's not obvious at all. A few I've identified:
- Campaign mailbox send limit
- account send limit
- weekly schedule
- if a account is in warmup
and there are still a few I haven't discovered because my campaign send about 20 mails/day, with 3 accounts, and the first 2 limits are at 50.

This creates a lot of mental load to figure out why your campaign is not delivering, a number somewhere that says how many email would it send, taking into account every limit and effect would be great. 

Edit: I realise now top left chart says I should be delivering 120 emails today, but it's missing something because it doesn't send as much.

meszmate:
Closes #606.

The campaign page promised a number the scheduler never intended to send: the sidebar meter summed every mailbox's configured cap client-side, and nothing named the mechanisms that decide a day's real volume.

**What this adds**

- `GET /campaigns/:id/send-plan`: today's projected sends, sent so far, what is still expected, a waterfall of every limit that removed something (the arithmetic adds up), the named bottleneck, the window, the lead supply counted through the campaign's own routing, a per-mailbox breakdown and the plan allowance. Derived through the scheduler's own gates on every read, never stored.
- A one-line **Today's sending plan** strip at the top of the campaign overview with a "Why" pane: the waterfall (each row links to where the limit is changed), leads, and each mailbox's day.
- The sidebar meter's denominator now comes from a `capacity_today` field on the dashboard payload, and the wizard estimate uses the same clamps and the plan allowance, so all three agree by construction.
- The compact Advisor strip folds to a summary line and its rows are single lines.
- `warmbly campaign plan` and `warmblyctl campaign plan`; docs, OpenAPI and skills updated.

**Behaviour change in the send path**

The campaign chain floored its next tick at the selected mailbox's whole min gap, so a three-mailbox pool sent at one mailbox's rate. The floor is now the pool's spacing; each mailbox's own gap is still enforced against its own last send.

**Verification**

`make lint`, `golangci-lint`, `gofmt`, web and docs typecheck and lint. New unit tests hold the plan's staged cap to `explainCap`. New live tests (lead supply, the per-sender ledger, an end-to-end plan that must add up, a paused campaign) pass against a database at the migration head, as do the existing routed-pairs and scheduler live suites.

## Summary by CodeRabbit

* **New Features**
  * Added a live “Today’s sending plan” view to campaign pages, showing projected sends, sent emails, remaining capacity, limiting factors, lead availability, sending windows, and mailbox details.
  * Added read-only API and CLI support for retrieving a campaign’s sending plan.
  * Dashboard sending-capacity metrics now reflect scheduler-applied limits.

* **Bug Fixes**
  * Corrected campaign and sidebar send totals to use actual scheduler capacity rather than configured mailbox caps.

* **Documentation**
  * Documented the new API endpoint, CLI command, and campaign sending-plan details.


meszmate:
@coderabbitai full review

